PR #111, opened 2026-08-08
SCS-039: moisture on the 2m grid, plus the SCS-018 positional read fix and the Esc chrome unification

TheCodingDad-TisonK:
> **Scope note:** this PR was opened on 2026-08-08 for the `getMoisture` positional-read fix alone. `development` is the trunk, so the SCS-039 build landed on top of it and this PR now carries **six commits**. Title and body updated to match what is actually here rather than what it was opened for.

## What is in this PR

| Commit | What |
|---|---|
| `330215f` | docs: SCS-018 positional read facade fix |
| `7620439` | fix(ui): unify Esc door chrome across all nine carriers |
| `b20f544` | **SCS-039** vendored value-map substrate + the quantisation law |
| `0ab019f` | **SCS-039** wire the value map behind engine detection |
| `172907f` | **SCS-039** conserved drainage on fresh height reads |
| `a97e217` | **SCS-039** MP delivery, release gate, degrade, instrument, seam |

**Suite: 325 assertions passed, 0 failed, across 12 files** (was 249/10 before SCS-039).

---

# SCS-039 / GRID-1: the ground gets one ruler

Moisture moves onto a 2 m value map, the same grain the rest of the soil suite measures by, instead of the shipped 10-40 m Lua cell store.

**Hosting is vendored per TysonK's ruling.** SCS carries its own machinery rather than registering a layer inside SoilFertilizer. Confirmed structurally while building: `SoilValueMaps.LAYER_DEFS` is a **static table with no registration API**, so there was never a door for SCS to register a seventh layer through. SF-hosted would have required SF to grow one first. SCS is sole writer; SF's six layers are never touched.

## Engine-absent is today, exactly

Every branch is gated on one `mapActive()` test. When it is false the shipped sparse-cell store is the whole system, bit for bit, including its relief pass, conserving drainage and packed persistence. **That invariant has its own bench**, because a regression in it is invisible on any machine where the map works, which is most of them.

## The quantisation floor is the law the build is shaped around

Moisture spans 0..1 across 254 raw steps, so **one raw step is ~0.0039 moisture** and an hourly weather delta is routinely smaller. A naive per-hour write floors to zero every time: **twenty four of them move the map by nothing** and the ground silently stops answering the weather.

Sub-step deltas accumulate through `quantiseDelta()` and only whole steps are spent, remainder carried, truncating toward zero in both directions so a drying field cannot oscillate. **The bench reproduces the broken behaviour first**, so anyone who later deletes the accumulator thinking it is an optimisation is told why it is not.

## Conserved drainage, on fresh height reads

Per the geometry audit's build law: terrain is deformable, so a cached relief picture drains water toward a hollow a player may have filled in weeks ago.

The maths is two terms and **each sums to zero on its own**, so conservation needs no correction pass to rescue it. An earlier draft used a sign-juggling relief bias plus a correction pass; it was replaced, because a conservation law you have to repair afterwards is one nobody can check by reading it.

## MP, gate, degrade, instrument

- **Delivery** is the six-layer precedent: eight rows per frame to a joining client, raw values run-length packed. Never one payload, since a 2048x2048 map is 4 MB.
- **`cs_grid_concordance` ships LOCKED and actually gates** — without the experimental opt-in the map never stands up.
- **The degrade is real**: a map that cannot load leaves no half-built object and does not re-probe; the per-field scalar rows carry the save.
- **`csMapStats`** reports the frame cost the brief left as its one open measurement.

## The SCS-037 seam

`hourlyUpdate` takes `elapsedHours`, defaulting to 1, arithmetically identical to what shipped. When SCS-037 replaces the edge detector with a real elapsed count it passes that count and nothing else changes.

## Two defects found and fixed in flight

1. **`readAverageOfPolygon` guarded on `getDensityMapModifierStats`, which has zero occurrences in the decompiled engine.** It would have returned nil forever and the field scalar would never have re-derived. Now guarded on the modifier, with `executeGet`'s real return order confirmed at the decompile: `(acc, numPixels, totalArea)`, and the divisor is `numPixels`.
2. **A field-level set did not reach the map**, so `csSetMoisture` and the sprayer path would have appeared to work then silently reverted on the next daily re-derive.

## Not verified in game

Nothing here has run in a live session. The maths, the delegation and the packing are proven on the bench; the engine plumbing is not. The system is release-gated off by default, which is the correct risk to carry at merge.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -51,3 +51,7 @@
 
 ## 2026-08-08 (Fred): SCS-018 per-cell moisture store (combined spatial-soil drop, head)
 - [x] Moisture is now a property of the ground, cell by cell, on the same grid SoilFertilizer uses (10m at 4x, 20m at 8x, 40m at 16x). Cells materialise where relief exceeds the threshold or water is applied; the field scalar stays the derived aggregate. Single write path, packed persistence (StateLedger SCHEMA 2 + XML), daily settle via the TimeGuard `simulation` flow class with a day-hook fallback, per-cell pivot/drip/sprayer water, Irrigate Now server-routed, and the RealisticWeather moisture unwind (RW stays a weather source only). Three offline benches green (relief sparsity, drainage conservation, real-scheduler catch-up). In-game observation owed (acceptance checks in the brief).
+
+## 2026-08-08 (Fred): SCS-018 positional read facade fixed
+- [x] The SCS-018 contract claimed getMoisture(fieldId, x, z) shipped, but the CropStressManager facade forwarded only fieldId, so SF-18's positional reads degraded to the field aggregate. SoilMoistureSystem already had the positional getter; the facade now forwards x, z. One-arg callers unchanged. Suite 162/0 across 9 files. Built and deployed. The SF-18 keystone (FS25_SoilFertilizer) consumes the positional read.
+
--- a/TODO.md
+++ b/TODO.md
@@ -58,4 +58,6 @@
 - [x] Daily TimeGuard `simulation` accrual + day-hook fallback + version-skew guard.
 - [x] RealisticWeather moisture unwind (2 writes + 2 step-asides removed; RW stays weather-only).
 - [x] Three offline benches (relief sparsity, drainage conservation, real-scheduler catch-up). Suite 162/0.
+- [x] SCS-018 positional read facade fixed: CropStressManager.getMoisture forwards x, z to SoilMoistureSystem (the 1-arg facade was the uncommitted gap; SF-18 consumes the positional read). Suite 162/0.
 - [~] In-game acceptance owed: hollow stays wetter over dry days; continuity on old saves; skip lands exactly; cell round-trip on both save paths; Irrigate Now survives broadcast on a client; RW map moisture evolves on our model.
+
--- a/main.lua
+++ b/main.lua
@@ -59,6 +59,7 @@ source(modDir .. "src/integrations/CropStressMasterHUDBridge.lua")
 source(modDir .. "src/events/CropStressSettingsSyncEvent.lua")
 source(modDir .. "src/events/CropStressMoistureInitEvent.lua")
 source(modDir .. "src/events/CropStressIrrigateNowEvent.lua")
+source(modDir .. "src/events/CropStressMoistureRowEvent.lua")
 
 -- Persistence
 source(modDir .. "src/SaveLoadHandler.lua")
@@ -269,6 +270,7 @@ Mission00.loadMission00Finished = Utils.appendedFunction(Mission00.loadMission00
         addConsoleCommand("csDebug",       "Toggle verbose debug logging",                                 "consoleToggleDebug", g_csManager)
         addConsoleCommand("csConsultant",  "Open the Crop Consultant dialog",                              "consoleConsultant",  g_csManager)
         addConsoleCommand("csRelease",     "Release gate: show STABLE vs experimental-LOCKED systems",     "consoleRelease",     g_csManager)
+        addConsoleCommand("csMapStats",    "Moisture value map: grain, settle cost, sync progress",       "consoleMapStats",    g_csManager)
     end
 end)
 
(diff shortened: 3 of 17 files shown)


TheCodingDad-TisonK:
Heads up for review: this PR has picked up the 2026-08-08 Esc door work since it was opened, so it now carries more than the title says.

- `7620439` **Esc door chrome unified across all nine carriers.** `RfPdaMenuPage.lua` had drifted into four versions. The door is built by whichever carrier loads first and load order is alphabetical, so DairyCore was winning and serving a chrome whose Soil footer was frozen at `{ Back, Help }`. Rotation Planner and Field Detail never reached the player, and Help rendered but did nothing because that copy called bare cross-mod globals with no resolver. MarketDynamics was a strict function-level superset and became the canonical base. All nine now ship one identical file, EOL-normalized hash `47142adce92f`.

Verified: parses clean under `luaparse` pinned to Lua 5.1 (control file rejected first), deployed zip re-opened after packing to confirm the hashes, and confirmed in game.

Pattern doc for future footer work: `Office Wizard/esc-rf-pda/CANONICAL-CHROME-AND-BUTTON-PATTERN.md`. The short version: a footer change is a nine-repo change, and footer buttons are Lua `menuButtonInfo`, never XML.

sasha-rf (approved):
Approving, and the first thing worth saying is about the PR itself rather than the code.

## THIS PR QUIETLY BECAME SOMETHING ELSE, AND THAT IS NOW FIXED

It was opened on 2026-08-08 for the getMoisture positional read fix. `development` is the trunk, so four SCS-039 commits and the Esc chrome unification landed on top of it and joined the same PR. Its title still described the small fix.

Anyone reviewing or merging would have believed they were taking a one line contract fix and would in fact have taken a whole new moisture storage substrate. Title and body now list all six commits. Caught before merge, not after.

**This is a standing hazard of a long lived development to main PR and it will happen again.** Worth somebody deciding whether these should be opened per feature and closed on merge, rather than one PR living on the branch indefinitely.

## SCS-039 ITSELF

**The quantisation floor is the part that makes this build worth trusting.** One raw step is about 0.0039 moisture and hourly weather deltas are smaller than that, so the naive version writes nothing, forever, and the ground silently stops responding to rain. The accumulator is not an optimisation and the bench says so by reproducing the broken behaviour first. If somebody deletes it later thinking it is a performance tweak, the test tells them exactly why they are wrong. That is the right way to leave a landmine warning.

**Engine absent is bit for bit today, and it has its own bench.** I pushed on this specifically. A regression in the fallback path is invisible on any machine where the map works, which is nearly all of them, and it would only ever be found by the one player whose setup cannot carry the map. Those players do not file good bug reports because nothing looks broken, it just quietly behaves differently. Guarding it with tests is the only honest way to ship this.

**Drainage was rewritten rather than patched.** The first version got the right answer through a correction pass. The replacement is two independently zero sum terms so conservation is visible by reading it. Throwing away working code because a reader cannot verify it is the correct call and I want it on the record that it was made.

**The two defects found in flight are both the good kind.** An engine call that does not exist anywhere in the decompile, and a field level set that would have appeared to work then reverted itself hours later. The second is nastier: `csSetMoisture` doing nothing is a bug report, `csSetMoisture` working and then undoing itself is a player losing trust in the whole mod.

## WHAT I AM NOT SIGNING OFF ON

**None of this has run in a live game.** The maths, the delegation and the row packing are proven on the bench. The engine plumbing, the map allocation, the region ops, the MP row walk, all unproven.

It ships release gated OFF, so a default install is untouched and that is the correct risk to carry at merge. But the unlock cannot happen on bench evidence, and `csMapStats` exists precisely so the frame cost question gets an answer rather than an opinion.

**The Esc chrome and getMoisture commits are older work I am not re-reviewing here**; they were already on the branch before SCS-039 arrived.

## PLAYER FACING READ

Nothing, today, by design. What a player eventually gets is soil moisture that varies across a field on the same grid everything else uses, instead of in 40 metre blocks. That is the difference between a hollow staying wet and a quarter of the field staying wet.

Approving for the one review rule. Tyson lands it.

sasha-rf:
﻿@TheCodingDad-TisonK verified and approved, ready to merge.

PR #112, opened 2026-08-10
SCS-037 the caught-up hour, and F154 the UsedPlus wear retirement

TheCodingDad-TisonK:
> **THIS PR CARRIES TWO ITEMS.** `development` is the trunk, so F154 landed on top of SCS-037 and joined this PR after it was opened and approved. Retitled and rebodied rather than left describing half its contents, per the hazard Fred reported on 2026-08-09. **Sasha's first approval covered `c890007` only and she has re-reviewed for `adeccbb`.**
>
> - `c890007` **SCS-037**, the caught-up hour
> - `adeccbb` **F154**, retire the UsedPlus wear bridge

---

# 1. SCS-037 — the caught-up hour

**The hourly tick is an EDGE DETECTOR.** It fires once when the hour key changes, however many hours the key jumped. A player sleeping three days got **ONE hour** of evaporation, rain, irrigation and stress; the other 71 were discarded on the line that computed the key. That is `F159`.

The delta the detector already holds is now read **before** the key is advanced and carried as a parameter to `onHourlyTick`.

## What multiplies, and what deliberately does not

| Consumer | Quantity |
|---|---|
| `SoilMoistureSystem:hourlyUpdate` | evaporation, rain gain, irrigation gain (seam already opened by SCS-039) |
| `CropStressModifier:hourlyUpdate` | stress accrual |
| `FinanceIntegration:chargeHourlyCosts` | irrigation running cost |

**The daily drainage settle is UNTOUCHED.** It rides its own day accrual and already honours `boundariesCrossed`; the one-clock rule says a quantity is charged by exactly one clock.

**The money now shares the water's clock, and that is not a side effect.** The moisture path already scaled irrigation *gain* by the elapsed count, so leaving the charge at one hour was handing a player 72 hours of free pumping.

## The cap, and the three ways a span is refused

`MAX_CATCHUP_HOURS = 168`, one in-game week. A ceiling rather than a tuning number: a week of unattended weather is already a total answer for a field, and an uncapped delta turns a corrupt or first-ever key into an unbounded simulation step.

`elapsedHoursFrom` returns **exactly 1** for the first tick ever (`lastHourKey` seeds to `-1`, so without this guard a fresh save would open with a full capped catch-up of weather it never had), for a key that did not advance, and for a key that moved backwards. **Every parameter defaults to 1, arithmetically identical to what shipped.**

## Round 2 ships INERT, and the reason is the contract rather than the feature

The brief reconstructs the rain switch per skipped day from SoilFertilizer's Water Record (SF-49). **`MaterialWetness:waterDaysInLast` exists and works but is not published** — the only path is `soilFertilityManager.soilSystem.materialWetness`, three internal field names deep, which is exactly the coupling SoilFertilizer's own delegate rule forbids.

So `getSkipRainHours` **probes for a manager-level delegate and answers nil until one exists**, which is round-1 behaviour exactly and the brief's own stated fallback. **The ask is one method:** `SoilFertilityManager:getWaterDaysInLast(days, throughDay)` returning `(count, known)`, nil-safe, neutral when the record is absent or the `ground_material` gate is closed.

---

# 2. F154 — retire the UsedPlus wear bridge

**The bridge read equipment condition out of UsedPlus DNA to scale irrigation flow, and it never returned anything.** DNA tracks vehicles; irrigation systems are placeables and never have an entry, so the read was always `0.0` and `wearFactor = 1.0 - wearLevel * 0.3` was a permanent multiply by one. **Today by accident, after this by design.**

The bridge itself was correctly built — runtime-detected, read-only, neutral when absent. **That is exactly why nobody noticed: not-needed and not-working are indistinguishable from outside a defensive fallback.**

**Removed:** `FinanceIntegration`'s `getUPAPI` resolver, `usedPlusActive`, the wear call inside the charge loop, `getEquipmentWearLevel` and `enableUsedPlusMode`; `IrrigationManager`'s `wearLevel` field, `wearFactor` at **both** rate sites, and `updateSystemWearLevel`; the FinanceIntegration arm of the detection in `CropStressManager`.

**NO RATE MOVES.** That is invariant 1 and the acceptance test. The bench pins each rate against the **pre-removal expression written out in its old shape** with `wearLevel` at its provable `0`, rather than against the new code.

`updateSystemWearLevel` is **deleted rather than kept** as a public setter for a future wear source: its only caller was this bridge, so keeping it would have made it a built-and-uncalled mechanism the moment the bridge went. **Nothing is repointed at AdvancedDamageSystem** — that is a new bridge with its own design pass, and repointing a dead probe at a live mod without one is how a fallback chain becomes load-bearing with nobody deciding.

## One deliberate deviation from the fix doc

**The doc names `FinanceIntegration` as the detection site's only consumer. It is not.** `UsedEquipmentMarketplace` is a second one, and the doc does not otherwise scope it. Retiring a whole listings feature is a wider decision than retiring a dead wear probe, **so the detection survives with the marketplace arm only, commented with why, and the marketplace's fate is reported rather than taken here.** `csStatus` now reports the marketplace's flag, which is UsedPlus's only remaining consumer in this mod.

This means **verification checklist item 3 (zero UsedPlus hits mod-wide) is deliberately not satisfied.** Items 1, 2, 4 and 5 are.

**README corrected:** that row claimed irrigation costs were tracked in the UsedPlus finance manager and that wear degraded pump output. Neither was ever true — costs have always gone through vanilla `addMoney`. `modDesc.xml` already only claimed the marketplace and needed no change.

---

# Bench

**381/0 across 14 files**, 56 new assertions (42 for SCS-037, 14 for F154). Every default and every rate is pinned against a hand-computed pre-change value rather than against itself.

# Owed before release

**The SCS-037 skip test.** Sleep or fast-forward ~72 hours on a save with a dry field and an active pivot, then `csStatus`: moisture drop, stress and money should each be about 72× a single hour, and the debug log now prints `Hourly tick complete (72h)`.

**F154 needs no in-game pass beyond that** — its acceptance is that nothing changed, and the bench measures it. Ships LOCKED at merge for SCS-037; F154 rides no locked system.

### DIFF
--- a/src/CropStressManager.lua
+++ b/src/CropStressManager.lua
@@ -109,6 +109,38 @@ CropStressManager.RWE_STRESS_MULTIPLIERS = {
     seed_growth_bonus   = 0.85,
 }
 
+-- SCS-037 THE CAUGHT-UP HOUR.
+-- The hourly tick is an EDGE DETECTOR: it fires once when the hour key changes,
+-- however many hours the key jumped. A player sleeping three days used to get ONE
+-- hour of evaporation, rain and irrigation, and the other 71 were discarded on the
+-- line that computed the key. The count is now carried as a parameter and every
+-- PER-HOUR term multiplies by it.
+--
+-- THE CAP is a deliberate ceiling on a single catch-up, not a tuning number: a
+-- week of unattended weather is already a total answer for a field, and an
+-- uncapped delta turns a corrupt or first-ever key into an unbounded simulation
+-- step. One in-game week.
+CropStressManager.MAX_CATCHUP_HOURS = 168
+
+--- Hours elapsed between two hour keys, as the hourly path should charge them.
+--- Pure and static so the bench can drive it without a mission.
+--- Returns 1 (today's behaviour exactly) for the first tick, for a key that did
+--- not advance, and for any key that moved backwards.
+---@param lastKey number|nil  the previously seen key, or -1/nil before the first
+---@param hourKey number      the key just observed
+---@return number hours       1 .. MAX_CATCHUP_HOURS
+function CropStressManager.elapsedHoursFrom(lastKey, hourKey)
+    local last = tonumber(lastKey)
+    local now  = tonumber(hourKey)
+    if last == nil or now == nil or last < 0 then return 1 end
+    local delta = now - last
+    if delta < 1 then return 1 end
+    if delta > CropStressManager.MAX_CATCHUP_HOURS then
+        return CropStressManager.MAX_CATCHUP_HOURS
+    end
+    return math.floor(delta)
+end
+
 
 function CropStressManager.new()
     local self = setmetatable({}, CropStressManager)
@@ -472,8 +504,11 @@ function CropStressManager:update(dt)
     local hourKey = day * 24 + hour
 
     if hourKey ~= self.lastHourKey then
+        -- SCS-037: the delta this detector already holds is the elapsed count.
+        -- Read it BEFORE the key is advanced, or it is gone.
+        local elapsedHours = CropStressManager.elapsedHoursFrom(self.lastHourKey, hourKey)
         self.lastHourKey = hourKey
-        self:onHourlyTick()
+        self:onHourlyTick(elapsedHours)
     end
 
     -- HUD frame update (handles auto-show, input response)
@@ -485,9 +520,24 @@ function CropStressManager:update(dt)
     end
 end
 
-function CropStressManager:onHourlyTick()
+--- SCS-037: `elapsedHours` is how many in-game hours this one tick stands for.
+--- It defaults to 1, which is arithmetically identical to what shipped, so every
+--- caller that does not care (the console heat simulator, any future direct call)
+--- keeps exactly today's behaviour.
+---
+--- WHAT MULTIPLIES AND WHAT DOES NOT. The three PER-HOUR consumers multiply:
+--- evaporation/rain/irrigation gain (SoilMoistureSystem), stress accrual
+--- (CropStressModifier) and irrigation running cost (FinanceIntegration).
+--- The DAILY DRAINAGE SETTLE DOES NOT: it rides its own day accrual and already
+--- honours `boundariesCrossed`, and the one-clock rule says a quantity is charged
+--- by exactly one clock. Weather is polled once and held: temperature and humidity
+--- are last-known sky across the whole span, which is the stated approximation.
+---@param elapsedHours number|nil  hours this tick covers (default 1)
+function CropStressManager:onHourlyTick(elapsedHours)
     -- Respect the player's master on/off toggle
     if not self.settings.enabled then return end
+
+    local hours = math.max(1, math.floor(tonumber(elapsedHours) or 1))
 
     -- Rebuild field map once per in-game day so newly bought/sold fields are tracked
     local env = g_currentMission and g_currentMission.environment
@@ -520,7 +570,10 @@ function CropStressManager:onHourlyTick()
         end
 
         -- 2c. Advance soil moisture simulation
-        self.soilSystem:hourlyUpdate(self.weatherIntegration)
+        -- SCS-037 round 2: when SoilFertilizer's Water Record is reachable, the
+        -- rain switch is reconstructed per skipped day rather than held at the
+        -- last-known sky. nil (the case today) means round-1 behaviour exactly.
+        self.soilSystem:hourlyUpdate(self.weatherIntegration, hours, self:getSkipRainHours(hours))
 
         -- 3. Apply RWE world-event stress multiplier (no-op when RWE not loaded)
         if self.rweManager then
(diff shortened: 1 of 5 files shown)


sasha-rf (approved):
Approved. Read the whole diff plus the two files it reaches into, and ran the suite: 367/0 across 13 files.

**The thing I checked hardest, because it is the claim the whole change rests on: is a normal hour really unchanged?** It is, and the bench earns that rather than asserting it. `soil.oneHourUnchanged` computes the expected value from `BASE_EVAP_RATE` and the loam params by hand and compares against the system, instead of comparing the system to itself with a different argument. That is the difference between a regression test and a tautology, and I have seen the tautology version pass on genuinely broken code.

**The water and money clock split is the real find here and it was already live.** SCS-039 multiplied irrigation gain by the span and left the charge at one hour, so since that merged a player sleeping three days with a pivot running has been getting 72 hours of water for one hour of money. Nobody filed it. It only appears if you go looking for the brief's third consumer, which is exactly what happened.

**Two observations. Neither blocks, both belong on the record.**

**1. SCS-037 WIDENS AN EXISTING ASYMMETRY ON THE TIME-GUARD-ABSENT PATH, and I want it written down before somebody rediscovers it as a bug report.** `SoilMoistureSystem:checkDayFallback` hardcodes `settleDaily(1)` regardless of how many days were crossed (`SoilMoistureSystem.lua:1073`), and its own comment says it accepts that limitation. That was survivable when evaporation also only ever charged one hour. After this change, a 72 hour skip WITHOUT Time Guard runs 72 hours of evaporation against ONE day of drainage redistribution. The drainage half is untouched by design and the brief's one-clock rule says so, so this is correct as merged. It is just that the gap between the two paths is now three times wider than the diff makes it look, and a player without Time Guard is the one standing in it. Worth a finding id.

**2. `known == 0` returning nil rather than zero is the right call and I want to say so out loud.** Zero rain hours is a claim that it was dry. Nil is an admission that we do not know, and only nil falls back to the honest approximation. A less careful version of this function would have returned zero and quietly stopped rain existing across every skip on any save where the record had not warmed up yet. The bench pins it (`round2.unknownIsNilNotZero`).

**On round 2 shipping inert: correct, and the reason is the one I would have raised anyway.** `waterDaysInLast` is reachable only through `soilFertilityManager.soilSystem.materialWetness`, which is three internal field names of another mod. That is the exact coupling I flagged on SF-52's getters last week and that SoilFertilizer rewrote as manager delegates to close. Probing for the delegate and answering nil until it exists is the same shape as the moisture band, and the bench guards it four ways including a throwing delegate. Building it the reachable way would have been faster and would have handed us a break on somebody else's rename.

**Nil safety and defaults checked in every direction.** `elapsedHoursFrom` refuses the first tick (`lastHourKey` seeds to `-1`), a key that did not move, and a key that moved backwards. `tonumber` guards a garbage span in all three consumers. `operationalCostPerHour` gained an `or 0`. The console heat simulator calls the one argument form and is untouched.

Sasha does not merge. #112 is Tyson's to land, and the in game skip test is still owed before the release gate opens.

sasha-rf:
﻿@TheCodingDad-TisonK verified and approved, ready to merge.

sasha-rf (approved):
Re-approving for `adeccbb`. My first approval covered `c890007` only and this PR has since gained a second item, so that approval was covering code it never saw. Same reason #808 was re-approved last week.

**F154 read in full, plus the fix doc it is built from. The suite is 381/0 across 14 files.**

**The acceptance test here is that nothing happens, which is the hardest kind to write honestly, and the bench gets it right.** `f154_usedplus_wear_retirement_test.lua` does not compare the new code to itself. It writes out the OLD expression, `flow * pressure * (1.0 - wearLevel * 0.3)`, with `wearLevel` at the value it was provably always pinned to, and compares the shipped rate against that. A test that had asserted `rate == flow * pressure` would have passed just as green while proving nothing about what players had before.

**Both rate sites moved together and I checked that specifically rather than trusting the message.** `activateSystem` and `applyOneTimeIrrigation` computed the same expression in two places, and a one-site fold is the defect class this project measures most. Zero hits on `wearFactor` and zero on `wearLevel` across the whole mod.

**Deleting `updateSystemWearLevel` instead of keeping it as a public setter is the right call and it is the one I would have argued for.** Its only caller was the bridge. Keeping it would have produced a built-and-uncalled mechanism in the same commit that removed one, which is the exact shape this suite already carries too many of.

**On the deviation, and I want to be clear that I agree with it.** The fix doc's invariant 3 says the detection goes with the consumers, and names FinanceIntegration as the only consumer. It is not one: `UsedEquipmentMarketplace` also arms off that site. The choice was between silently killing a 150 line listings feature that no doc scoped, or leaving the detection alive for its one real remaining consumer and saying so. **Leaving it and reporting it is correct.** A dead probe with no consumer is the thing invariant 3 forbids, and this one still has a consumer. The comment at the site explains it, so the next reader does not have to rediscover the gap.

**What I want on the record as a consequence:** checklist item 3, zero UsedPlus hits mod wide, is now deliberately unmet and will stay unmet until somebody rules on the marketplace. That is a decision with a name on it rather than an oversight, but it should not sit unowned. Given Tyson's finding that UsedPlus is no longer maintained, a listings feature that can only ever register with a frozen mod is worth a design answer soon.

**The README correction matters more than its size.** That row promised players two things, that irrigation costs were tracked in the UsedPlus finance manager and that wear degraded pump output. Neither was ever true, and the code's own comment said costs go through vanilla `addMoney`. We were shipping a false claim before this commit, not because of it.

**No behaviour change found anywhere I looked.** The charge loop dropped its unused `id` binding and still charges the same money, and SCS-037's span multiply survives it intact.

Sasha does not merge. #112 is Tyson's to land, and the skip test is still owed.

sasha-rf:
﻿@TheCodingDad-TisonK verified and approved, ready to merge.

PR #114, opened 2026-08-10
chore: correct the stale SCS-037 round-2 inert comment (round 2 is live)

TheCodingDad-TisonK:
Correct the stale comment: SCS-037 round 2 is live, not inert.

## Why

The round-2 rain-switch reconstruction in `getSkipRainHours` was written inert on purpose, probing for a manager-level Water Record delegate that did not exist yet. SoilFertilizer shipped that delegate in PR #811, so the probe now succeeds and round 2 is live. The code needed no change, it was already correct and its bench already covered the live path. What was left behind was the narrative: the comment block still said "IT IS INERT TODAY" and the bench still framed itself as the inertness guard.

## What

- Corrected the `SCS-037 ROUND 2` comment block in `CropStressManager.lua` to state the delegate is published and round 2 is live, and that the nil paths are kept because they remain the honest answer when the record is absent or the ground_material gate is closed.
- Corrected the matching "inertness guard" framing in `scs037_caught_up_hour_test.lua` to the nil-safety guard it actually is.
- Removed the em dashes in the rewritten comment block.

## Verification

Comment-only change. Suite 381/0 across 14 files; Lua 5.1 syntax clean. Not a build or gameplay change, nothing to deploy.

The 72-hour in-game skip test is still owed and is the only remaining item on the SCS-037 chain.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -55,3 +55,8 @@
 ## 2026-08-08 (Fred): SCS-018 positional read facade fixed
 - [x] The SCS-018 contract claimed getMoisture(fieldId, x, z) shipped, but the CropStressManager facade forwarded only fieldId, so SF-18's positional reads degraded to the field aggregate. SoilMoistureSystem already had the positional getter; the facade now forwards x, z. One-arg callers unchanged. Suite 162/0 across 9 files. Built and deployed. The SF-18 keystone (FS25_SoilFertilizer) consumes the positional read.
 
+
+## 2026-08-10 (Fred): SCS-037 round 2 is live - the rain switch across a skip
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):
Approve. Comment-only correction, reviewed against the merged state.

What checks out:

- The claim in the PR is verifiable: SoilFertilizer PR #811 shipped `getWaterDaysInLast` on the manager, and `getSkipRainHours` at `CropStressManager.lua:650` probes for exactly that method. So the "IT IS INERT TODAY" block was stale the moment #811 landed, and this correction is accurate, not cosmetic.
- The rewrite states round 2 is live and keeps the nil paths, which are still the honest answer when the record is absent or the ground_material gate is closed. The code was not touched, matching the brief that the consumer already probed and its bench already covered the live path.
- The bench framing change matches reality: the block is a nil-safety guard, not an inertness guard, and the assertions still pin every unknown path.
- Em dashes in the rewritten comment block are removed. The diff is em-dash clean.
- Suite 381/0 across 14 files, syntax clean. No build or gameplay change, so nothing to deploy.

The 72-hour in-game skip test remains the only open item on the SCS-037 chain, and it is flagged in the PR body and the docs, not hidden.

No release implied by this review.

PR #115, opened 2026-08-10
fix(gui): IrrigationScheduleDialog crashes on open after F154

TheCodingDad-TisonK:
Fix the irrigation schedule dialog crashing on open, a regression from F154.

## Why

F154 retired the UsedPlus wear bridge and deleted `updateSystemWearLevel`, so `system.wearLevel` is now nil on every irrigation system (the F154 bench pins this: `s.wearLevel == nil`). The schedule dialog still read that field in `updatePerformance` - the flow-rate and wear readouts did `system.wearLevel * 0.3` and `system.wearLevel * 100`. That threw `attempt to perform arithmetic (mul) on nil` during `onOpen`, after the dialog was already shown, so it stayed visible with input never initialised: it could be neither interacted with nor closed. Found in-game on the 72h skip test session.

## What

- `IrrigationScheduleDialog:updatePerformance` reads wear as `system.wearLevel or 0`. Per F154 the wear factor is a permanent multiply by one after the retirement, so nil and zero are the same number - this changes no player-observable behaviour, which is F154's invariant 1.
- The F154 bench gains a dialog-consumer assertion under the same invariant: a system with nil wear yields the pre-removal rate, not a throw.

## Verification

Suite 382/0 across 14 files; Lua 5.1 syntax clean. Built and deployed. Not yet opened in-game with the fixed zip (the crashing session was ended).


sasha-rf (approved):
Approve. Reviewed against source, this is a real regression fix, not a workaround.

What checks out:

- Root cause verified: F154 deleted `updateSystemWearLevel`, so `system.wearLevel` is nil on every system, which the F154 bench itself pins (`gone.noWearField` asserts `s.wearLevel == nil`). The schedule dialog read that field in `updatePerformance` for both the flow-rate and wear readouts, throwing on nil during `onOpen` after the dialog was already shown, so input never initialised. The log line matches exactly.
- The fix honors F154's invariant 1: the wear factor is a permanent multiply by one after the retirement, so `wearLevel or 0` produces the identical rate. No player-observable behaviour changes.
- The bench addition sits under the right invariant: a system with nil wear yields the pre-removal rate rather than a throw, pinning the dialog consumer against the same hand-computed number the F154 test uses elsewhere.
- Suite 382/0 across 14 files, syntax clean, built and deployed. The zip carries the guard.

Not yet verified in game: the crashing session was ended before the fixed zip could be opened. Flagged in the PR body, and the next launch should confirm the dialog interacts and closes normally.

No release implied by this review.

sasha-rf (approved):
Re-approve after the split. This PR is now exactly two commits: the F154 dialog crash fix and its own docs entry. The SCS-037 verification docs moved to their own branch per rule zero.

Reviewed the current head (33c65c0):

- `IrrigationScheduleDialog:updatePerformance` reads `system.wearLevel or 0`, so the dialog no longer throws on open after F154 deleted the wear setter. Matches F154's invariant 1: wear was always a multiply by one, so the rate is unchanged.
- The F154 bench gains the dialog-consumer assertion under the same invariant. Suite 382/0, syntax clean, built and deployed, and confirmed in-game: the dialog loaded and set pivot 516's schedule.

No release implied by this review.

sasha-rf (approved):
Reviewing this SCS PR (directed by Tyson for the suite release).

**#115:** the F154 wear-nil dialog crash fix. Resolved against current development: development already carries the newer fix for the same crash (nil guards + the wear row hidden per Samantha's ruling), so this PR's Lua change was superseded by the merged version; the merge keeps development's version and the branch carries the remaining doc/test content cleanly.

**#116:** the SCS-037 caught-up hour in-game verification docs. Clean.

Approved. No em dashes in this review.

PR #119, opened 2026-08-12
feat(SCS-021): the drying model learns what humidity is

TheCodingDad-TisonK:
# feat(SCS-021): the drying model learns what humidity is

In one farmer's sentence: a hot dry wind takes the moisture out of a field
faster than a hot still muggy day, and today the simulation cannot tell those
two days apart because the humidity probe returns a constant.

## What ships (one file, three functions plus constants, no new store)

- `WeatherIntegration.computeVPDMultiplier(tempC, humidityFrac)`: Tetens SVP,
  `vpd = svp * (1 - humidity)`, `mod = (vpd / VPD_REF) ^ 0.40`, clamped to
  0.40 - 2.20 AFTER the exponent and BEFORE any composition.
- `getHumidity()` reads WeatherGuard FIRST (`getCurrentSky().humidity` +
  `humidityDefaulted`, table fields never positional), normalizes a percent
  read to a fraction exactly once, and now returns `(humidityFrac, defaulted)`.
  A nil sky or flag falls through to the existing chain; the tail is
  `return 0.5, true`, never a bare value.
- `update()` unpacks both returns; `currentHumidityDefaulted` initializes true
  in `new()` and is recomputed every tick, never persisted.
- `getHourlyEvapMultiplier()`: the honesty guard first - when humidity is
  defaulted, this returns today's formula bit for bit. Only with a live read
  does the VPD term replace the temperature-only linear term, on the same
  season and rain frame.
- `getMoistureForecast()`: the second copy dies. The hand-computed tempModMean
  is replaced by the same helper fed the drying window whole
  (`getForecastHumidity(day)` from WeatherGuard). A nil from either keeps the
  linear term for that day, silently.
- `csStatus`'s sky line appends `(humidity defaulted)` when the flag is true.
  No log output on the fallback path.

## What this is not

Not a fix for F159 (its own ruled member). Not the RealisticWeather branch
removal (SCS-019). Not a UI change (both inputs already render). Not a wind or
cloud model.

## Bench

- `SCS-021-drying_vpd_spec_test.lua`, 31 assertions: the VPD math at the
  reference, the honesty guard (defaulted = legacy formula bit for bit), the
  WeatherGuard-first read and its normalized/defaulted returns, the forecast's
  drying window, and the nil-keeps-linear-term path.
- Full suite green: 413 assertions, 0 failed across 15 files.

## In-game (a person's job, the bench cannot see it)

On a save with WeatherGuard active, `csStatus` shows no `(humidity defaulted)`
and the drying multiplier visibly differs between a humid and a dry day at the
same temperature. With WeatherGuard removed, drying behaves exactly as the
previous release.

26 languages: no new player-facing strings (the debug string is console-only),
so no translation files change.


sasha-rf (approved):
Approved. SCS-021 drying VPD. WeatherGuard's live humidity (with its honest defaulted flag) feeds a VPD multiplier in the hourly evap term and the forecast's drying window; absent real data, drying is today's behaviour bit for bit. Suite 413/0. No new player strings.

PR #120, opened 2026-08-12
feat(SCS-036): publish the field soil type

TheCodingDad-TisonK:
# feat(SCS-036): publish the field soil type

SoilFertilizer has been asking SeasonalCropStress what kind of ground a
field is since the day its drying model was written, and has never once got
an answer. The probe is correct, guarded and pcall-wrapped; the function it
calls has never existed. This publishes it. Four small edits in one mod, no
new state, no new wire payload, no new setting, and one shipped consumer
starts working with no change on its side.

## The four edits

1. **The backfill**, `SoilMoistureSystem:enumerateFields`: a record that
   EXISTS but has no `soilType` gets one detected and written. The field
   object is already in scope from the same loop. Idempotent: one detection
   per field per peer per session. The return contract's count now includes
   backfilled records, so a client that repairs every field and creates none
   reports the right work instead of 0.
2. + 3. **Both wire handlers drop the placeholder**, `CropStressMoistureInitEvent`
   and `CropStressNetworkSyncBridge`: `{ moisture = entry.moisture, soilType = "loamy" }`
   becomes `{ moisture = entry.moisture }`. The reason is structural:
   "loamy" the placeholder and "loamy" the real answer are the same string,
   so no code can tell them apart. An absent key is detectable. It is safe on
   the day it ships - the hourly loop already lands missing classes on loam,
   so the join window behaves identically to today and the first rebuild
   makes the value right. Both files, because which handler governs a session
   depends on whether NetworkSync is installed.
4. **The getter**, `CropStressManager:getFieldSoilType(fieldId)`, sited in the
   companion read API block after `getYieldKeepFactor` with the
   DERIVED-FROM-FIELD-ID warning header (it is a hash of the farmland id, not
   a terrain measurement). Pull-only: reads `soilSystem.fieldData[fieldId].soilType`,
   nil on a missing system or record, writes nothing, caches nothing. Named
   `getFieldSoilType` because that is the exact spelling the shipped SF probe
   calls; the wrong spelling would leave the probe still failing. SoilFertilizer
   is added to the companion list in the block header.

## Integration

Consumes nothing new. Provides: SoilFertilizer's `MaterialWetness:soilClassFor`
(shipped, probing) now resolves, and the spatial nutrients build (briefed,
unbuilt) uses the same read as its tier-0 texture baseline. An unknown class
returns nil, never a substitute. No money path, no new persisted state, no new
wire payload.

## Bench

- `SCS-036-soil_type_publication_spec_test.lua`, 12 assertions: pull-only
  getter with nil on every missing direction, the exact getter name locked,
  the backfill repairing old records and counting them, idempotence, and the
  loam fallback that keeps the join window unchanged.
- Full suite green: 394 assertions, 0 failed across 15 files.

## In-game acceptance (owed at unlock, not before)

1. Join a dedicated server as a client, then confirm that after the first
   daily rollover a field's soil class matches the host's.
2. Confirm nothing regresses in the join window, when the class is briefly
   nil and the fallback carries it.

No new player-facing strings, so no translation files change.


sasha-rf (approved):
Approved. SCS-036 soil-type publication. Four edits: the backfill, both wire handlers dropping the loamy placeholder, and getFieldSoilType on the companion read API. Pull-only, nil-never-substitute; SF's existing probe starts succeeding with no change. Suite 394/0.

PR #121, opened 2026-08-12
feat(SCS-038): irrigation costs what it draws

TheCodingDad-TisonK:
# feat(SCS-038): irrigation costs what it draws

Irrigation's operational cost now varies with the water it actually draws.
The flat `operationalCostPerHour` stays as the XML number at full pressure;
a distant or low-pressure source prices a higher effective hourly cost,
because less water moves for the same pump time.

## What ships

- `IrrigationManager:getEffectiveCostPerHour(system)`: returns
  `operationalCostPerHour / pressureMultiplier`. nil when there is no source
  or no pressure, which already means no run and no charge (never a number
  that would charge for nothing).
- The ROUND-2 LIFT term, designed-in and NEUTRAL: `effCost = base / pressure *
  (1 + LIFT_COEFF * max(0, pivotY - sourceY) / 10)`. `LIFT_COEFF` is an XML
  balance value (`liftCoeff` on both placeable configs) defaulting to 0.0, so
  this is bit-for-bit round-1 behaviour until the ratio pass tunes it. Heights
  are read once at registration (pivot and source Y stored on their records),
  so the term is cheap at charge time.
- `FinanceIntegration:chargeHourlyCosts` charges the priced rate through the
  getter, falling back to the flat per-hour number when the getter is absent
  or nil, so an unregistered edge never crashes the charge.

## Neutral beside a pump

At full pressure (multiplier 1.0) the effective rate is exactly the XML
number: 15 for a pivot, 8 for a drip line. At range (pressure 0.7) a 15-per-hour
pivot charges 21.43 per effective hour. SCS-024's held design consumes the same
getter when it builds; neither edits the other.

## Bench

- `SCS-038-priced_draw_spec_test.lua`, 13 assertions: base / pressure at
  neutral and range, nil on no-source and zero-pressure, the LIFT term neutral
  at 0.0 and live above it (including the downhill max(0,...) kill), and the
  charge line using the priced draw with a flat fallback.
- Full suite green: 395 assertions, 0 failed across 15 files.

## In-game (a person's job, the bench cannot see it)

Beside a pump the charge is unchanged; a pivot fed from a distant pump charges
more per effective hour. The XML `liftCoeff` stays 0.0 so nothing visibly moves
until the ratio pass tunes it.

26 languages: no new player-facing strings, so no translation files change.


sasha-rf (approved):
Approved. SCS-038 priced draw. getEffectiveCostPerHour prices what a system draws (base/pressure, nil when no source), with the neutral LIFT term designed in; the charge line falls back flat on any nil. Neutral beside a pump. Suite 395/0.

PR #122, opened 2026-08-12
fix(SCS-043): reel N binding, pump turn-on, gate diagnostics, hose ground clamp

TheCodingDad-TisonK:
Found by Tyson's in-game pass on the irrigator play kit (Rainstar T32, Aggregat EMP, sprinkler cart), 2026-08-12.

**Original symptom:** connecting the pump hose deletes the N binding for automatic hose retraction. Disconnecting brings N back, but pressing it then warns that the motor pump hose must be connected. Unwinnable in both directions.

Restoring the binding exposed three further defects behind it, each of which had been unreachable while the control itself was missing. This PR carries both passes.

## Commits

| | |
|---|---|
| `b651939` | `fix(SCS-043): restore reel N binding while the pump hose is connected` |
| `3cb49be` | `fix(SCS-043): turn the pump on, make every reel gate speak, keep the hose above ground` |

**Retitled and rebodied 2026-08-12** when the second commit landed, so the title describes everything inside. No approval had been given at that point, so none needed refreshing.

## Pass one: the binding

**1. N was handed to a seat that never took it.** `rainstarAutoReel.lua:1204` deliberately skipped registering `RWSM53_REEL_TOGGLE` whenever the hose was connected, on the stated intent that N belongs to the pump. That action is registered in exactly one place in the whole mod: the block behind that guard. Connecting the hose simply deleted the binding.

The Rainstar always registers N now. The double-toggle the guard was written against cannot happen, because both surfaces are vehicle action events and a player is active for input in one vehicle at a time.

**2. N came back only when it could not work.** `rarStart()` gates on `rarIsPumpConnected`, which reads `vehicle.rwsmVirtualPump` - cleared by the disconnect that just restored the binding.

`ScsPumpHoseConnection` now registers N as well, proxying to `RWSM53AutoReel.toggleReel`, a public helper that shipped with the vendored file and had zero callers. Hidden until a Rainstar is actually on the hose; label follows start/stop state.

## Pass two: what was behind it

**3. The pump could never be turned on, by any path.** `Sprayer:getCanBeTurnedOn` returns false unconditionally when `allowsSpraying` is false (`Sprayer.lua:696`), and `Aggregat.xml:167` sets exactly that. The flag is correct - this is a water pump and must not behave as a sprayer - but it also walls off the vanilla turn-on binding, so no build of this mod could ever open the reel's pump-running gate. `log.txt` read `running=false` across every session on record.

`TurnOnVehicle:setIsTurnedOn` carries no precondition of its own (`TurnOnVehicle.lua:282`); `getCanBeTurnedOn` is a guard the vanilla *action handler* applies. The walk-up start now waits for the motor to finish cranking and sets the state directly. The XML already anticipated this: `turnOffIfNotAllowed="false"` is the flag that stops the game reverting a turn-on that is not "allowed".

`phIsPumpRunning` also no longer accepts a running diesel as a running pump. That fallback made the walk-up prompt read "Stop pump" the moment the engine caught, so the player pressed what looked like a stop and got an engine shutdown, because no pump had ever started.

**4. All six `rarStart` gates refused silently**, the rollover check included, which left the player with a bare blinking warning at best and the log with nothing at all. Each gate now names itself and the state it read, the rollover one with its measured up-vector values, and both N surfaces announce the press so a refusal can be told apart from a key that never fired.

**5. The supply hose was drawn below the terrain.** `updateHosePose` sagged a straight line between two couplings that both sit low on their machines, with no ground check anywhere. Interior points now clamp to terrain + 8 cm; the two ends keep their exact coupling positions so a joint sampling below terrain height on a slope cannot tear the hose visual off the machine it is plugged into.

## Verification

- Lua 5.1 syntax gate green, 44 files, on every commit. Control-tested with a deliberate break first, because that checker has passed broken files before.
- Engine behaviour verified against the decompile at `D:\FS25_Decoded`, not from memory: `Sprayer.lua:696`, `TurnOnVehicle.lua:282/343`, `Motorized.lua:2499`, `Vehicle.lua:2067`.
- In-game, after the turn-on fix: `connected=true running=true pressure=8.5 bar radius=33 m flow=34.0 m3/h`, followed by `[CropStress] ScsPumpHose: pump turned on, reel gate is open`.
- Retraction observed working by Tyson. Not captured in a log, since that session's log.txt was overwritten by the next launch.
- No file overlap with #121 (SCS-038).

## Filed separately, not in this PR

- **`src/ScsRainstarPumpPower.lua` does not exist** but `modDesc.xml:217` declares it, so `vehicle type 'rwsmPlayRainstar' has unknown specialization` on every load. **Resolved by Claude(A) 2026-08-12: neither write nor drop.** Wizard's local Crop Stress clone carries the file, and a PR of it closes the load error; writing it here would duplicate existing work, and dropping the declaration would delete design intent that went missing in the Esc handoff. Nothing for this PR to do.
- **Reel position does not persist.** `initSpecialization` registers `vehicles.vehicle(?).rwsmAutoReel#animationTime`, but mod specializations are handed a namespaced key and the write is rejected every save. FAST TRACK, its own branch next.
- **The Schiene falls through the world**, resetting to terrain level every twenty seconds or so through the whole test window. Unexplained, untouched by anything here, and its own defect.
- Every hose connect reloads `rwsmPumpHoseTemplate.i3d` at about 45 ms because disconnect deletes it. **Deferred by Claude(A) 2026-08-12 (item 7)**: a real fix, but it ranks behind the load error and the open defect PRs on this mod. Raise again once this PR and the pump-power file are closed.

### DIFF
--- a/src/ScsPumpHoseConnection.lua
+++ b/src/ScsPumpHoseConnection.lua
@@ -117,6 +117,19 @@ local function phIsPumpRunning(pump)
     return false
 end
 
+-- The connected Rainstar, but only when it actually carries the reel spec.
+local function phGetReelPartner(pump)
+    local spec = pump ~= nil and pump.spec_scsPumpHose or nil
+    if spec == nil or not spec.connected then
+        return nil
+    end
+    local partner = spec.partner
+    if partner == nil or partner.RAR == nil then
+        return nil
+    end
+    return partner
+end
+
 local function phWaterFillType()
     if g_fillTypeManager ~= nil then
         local idx = g_fillTypeManager:getFillTypeIndexByName("WATER")
@@ -148,6 +161,7 @@ function ScsPumpHoseConnection.registerEventListeners(vehicleType)
     SpecializationUtil.registerEventListener(vehicleType, "onDelete", ScsPumpHoseConnection)
     SpecializationUtil.registerEventListener(vehicleType, "onUpdate", ScsPumpHoseConnection)
     SpecializationUtil.registerEventListener(vehicleType, "onUpdateTick", ScsPumpHoseConnection)
+    SpecializationUtil.registerEventListener(vehicleType, "onRegisterActionEvents", ScsPumpHoseConnection)
 end
 
 function ScsPumpHoseConnection:onLoad(savegame)
@@ -168,6 +182,11 @@ function ScsPumpHoseConnection:onLoad(savegame)
     spec.hoseTemplate = nil
     spec.hoseSegments = {}
     spec.candidate = nil
+    spec.reelActionEvents = {}
+    -- Motor start is asynchronous: getIsPowered() is still false in the frame
+    -- tryStartMotor is called, so the turn-on has to be retried until it takes.
+    spec.pendingTurnOn = false
+    spec.pendingTurnOnTimer = 0
 
     -- Soft path for vendored parking brake / COBD (RPC.virtualHoseConnected).
     if self.RPC == nil then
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):
Approved. SCS-043 reel N binding. Three gates closed in order: the Rainstar always registers N, the pump hose connection proxies it, and the pump turn-on retries until powered. Suite green; deployed zip carries it. The two out-of-scope findings are correctly routed to design (PIPELINE) and their own FAST TRACK branch.

PR #123, opened 2026-08-12
fix: add missing ScsRainstarPumpPower file (Esc handoff)

WizardlyPayload:
﻿## Summary
- Adds `src/ScsRainstarPumpPower.lua` that `modDesc` already declares for the Rainstar play vehicle.
- Closes the live Can't load resource / unknown specialization load error Fred logged during the irrigator pass.
- **Not fully finished:** this restores the missing specialization file only. The EMP-powered reel path still depends on hose-connection helpers (for example `getIsPumpRunning`) that are not on `development` yet. Do not treat this PR as a complete irrigator power story.

## Test plan
- [ ] Load Crop Stress with Rainstar play kit; confirm no Can't load resource for ScsRainstarPumpPower
- [ ] Confirm vehicle type rwsmPlayRainstar no longer reports unknown specialization
- [ ] Note: full EMP-without-tractor spray may still need the follow-up hose helpers before it works end to end

### DIFF
--- a/src/ScsRainstarPumpPower.lua
+++ b/src/ScsRainstarPumpPower.lua
@@ -1,0 +1,86 @@
+-- =========================================================
+-- ScsRainstarPumpPower
+-- BUILD 22:25 - the EMP pump is the Rainstar's power source.
+--
+-- Why the reel demanded a tractor
+-- ------------------------------
+-- Rainstar's <attachable> block carries no <power> element, and the schema default
+-- for vehicle.attachable.power#requiresExternalPower is TRUE. So unattached, with no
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Approved. Adds the missing ScsRainstarPumpPower.lua that modDesc already declares, closing the Can't load resource / unknown specialization load error from Fred's irrigator pass. getIsPowered overwrite gates on a live pump hose and defers to superFunc otherwise, so the tractor path is unchanged. The PR's own note that the full EMP path still needs hose helpers is accurate and correct. Verify the Rainstar loads without the error in-game.

PR #124, opened 2026-08-12
fix(scs): leftover hose/Rainstar/Esc nits (no farm patches)

WizardlyPayload:
fix(scs): leftover hose/Rainstar/Esc nits (no farm patches)


sasha-rf (approved):
Approved. Leftover share-wave work per Wizard's closeout: Esc chrome deltas, HUD bridges, SettingsHub guard, ProStaff nits. Matches the ASH-INBOX closeout list; farm patches correctly excluded.

PR #126, opened 2026-08-12
fix: remove source() of the unshipped FarmPatchUtil.lua

TheCodingDad-TisonK:
# fix: remove source() of the unshipped FarmPatchUtil.lua

Wizard's leftover PR kept source(modDir .. src/ui/FarmPatchUtil.lua) while excluding the farm-patch code from the share, so every load errors with Can't load resource FarmPatchUtil.lua. Farm patches are deliberately local (unpaid playtest). CsRfPdaGuest references it nil-guarded, so dropping the source call is safe.


sasha-rf (approved):
Approved. Restores the tested pre-leftover version of the Esc chrome files (the drifted leftover tree broke the door UI) and drops the source() of the unshipped FarmPatchUtil. Verified the chrome now matches the parent byte-for-byte; F155 translations preserved.

PR #127, opened 2026-08-12
Esc Crop Stress: restore Help, hide consultant SPACE chip

WizardlyPayload:
## Summary
- Restore Esc Crop Stress footer Help (opens Crop Stress help dialog).
- Hide the on-page SPACE Crop consultant activate chip.
- Keep Agronomist / Pivot / field table as-is.
- Nine-door sync of the Crop Stress Help footer branch on Esc door carriers.

## Test plan
- [ ] Full client restart
- [ ] Esc Realistic Farming to Crop Stress
- [ ] Footer shows Back + Help; Help opens Crop Stress help dialog
- [ ] No SPACE Crop consultant chip
- [ ] Agronomist and Pivot still visible

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.67</version>
+    <version>1.2.5.69</version>
 
     <title>
         <en>Seasonal Crop Stress &amp; Irrigation Manager</en>
--- a/src/ui/CsRfPdaGuest.lua
+++ b/src/ui/CsRfPdaGuest.lua
@@ -1161,24 +1161,36 @@ function CsRfPdaGuest.onOpenSchedule(container)
     end
 end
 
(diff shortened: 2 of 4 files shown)


sasha-rf (approved):
﻿Approved. This is the guest-side half of the Esc Help restore, and it is correct.

Verified:
- UTF-16 guard clean: all four changed files are plain UTF-8, no BOM, no NUL bytes.
- modDesc.xml: version bumped to 1.2.5.69.
- src/ui/CsRfPdaGuest.lua: `onOpenHelp` is defined with the same pcall-guarded pattern as its siblings (onOpenConsultant / onPaintConsultant). `CsDialogLoader` is nil-guarded before use, `CsDialogLoader.lua` and `CsHelpDialog.lua` both exist in the tree, and `refocusPivotAfterDialog` is a local defined above the new handler. The SPACE consultant chip is now hidden via setVisible(false) rather than being repainted, and `onOpenHelp` is registered in the module def.
- src/ui/RfEscModules.lua and src/ui/RfPdaMenuPage.lua: identical to the changes already merged in the other suite repos.

This is the actual provider for the host-side soft-detection added everywhere else. Merging completes the wave.

PR #128, opened 2026-08-14
fix(F158): the irrigation water bill goes to the system's owner, resolved at charge time

TheCodingDad-TisonK:
F158, the irrigation water bill. On a dedicated server the running cost resolved the farm from the local player, and a dedicated server has no local player, so no irrigation system was billed all season for any farm. On a listen server every system on the map was billed to the host, so the farmer paid for his neighbours' pivots. The pivot's owner was never consulted anywhere in the path.

The fix, exactly the way the base game charges water (PlaceableHusbandryWater with self:getOwnerFarmId()):
- registerIrrigationSystem now holds the owning placeable on the system record; deregister removes the record when the placeable goes, so the reference cannot dangle.
- The bill is resolved at charge time from system.placeable:getOwnerFarmId(), never stored, so a placeable that changes hands is billed to its new owner.
- deductFundsVanilla takes the farm it is billing and never derives one from the viewer; an unresolvable owner or the spectator farm is skipped, never billed.

This is the defect Arisanni certified on 2026-08-12 (the third hole in the water bill, the one that bills nobody on a dedicated server) and named the fix direction for. F159, the skipped-hours charge, was already closed in shipped code; this closes F158.

Verified: 12 new assertions (scs158_dedi_farm_resolution_test.lua): per-system owner charges on a dedi, the unresolvable-owner skip, and the deduction charging the given farm and never the viewer. The existing charge-path tests were updated for the new system shape. Full suite 448/0, syntax and lint clean. Built and deployed as 1.2.5.70. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -69,3 +69,8 @@
 - [x] F154 retired the UsedPlus wear bridge and deleted updateSystemWearLevel, leaving system.wearLevel nil on every system. The schedule dialog read that field in updatePerformance and threw on nil during onOpen, leaving the dialog visible with input never initialised (could not be interacted with or closed). Found in-game on the 72h skip test.
 - [x] Fixed by reading wear as system.wearLevel or 0. Per F154 the wear factor is a permanent multiply by one after the retirement, so nil and zero are the same number and no player-observable behaviour changes (F154 invariant 1). F154 bench gains the dialog-consumer assertion. Suite 382/0, syntax clean, built and deployed.
 - [~] Re-open the dialog in-game on the fixed zip to confirm it interacts and closes normally.
+
+## 2026-08-14 (Fred): F158 - the irrigation water bill goes to the system's owner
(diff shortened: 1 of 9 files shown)


sasha-rf (approved):
Approved. F158, the irrigation water bill now goes where the water actually belongs.

The running cost resolved the farm from the local player, so a dedicated server billed no system all season (no local player) and a listen server billed every system on the map to the host. The bill now resolves each system's owner from its placeable at charge time, the exact shape the base game uses for water (PlaceableHusbandryWater with self:getOwnerFarmId). registerIrrigationSystem holds the placeable on the record, deregister removes it when the placeable goes, and the owner is never stored, so a pivot that changes hands is billed to its new owner.

The deduction takes the farm it is billing and never derives one from the viewer; an unresolvable owner or the spectator farm is skipped rather than billed to nobody or to farm 0. Bench is 12 new assertions covering the per-system owner charges on a dedi, the skip, and the deduction charging the given farm. The existing charge-path tests were updated for the new system shape. Full suite 448/0. Not yet seen in a running game; the two-farm dedi billing pass is the owed in-game check.

PR #129, opened 2026-08-14
fix(F160): the weekly irrigation schedule honours the picked days, derived from the monotonic day

TheCodingDad-TisonK:
F160, the weekly irrigation schedule ran seven days on a default save. The day-of-week index was read from env.currentDayInPeriod, which the base game computes as (currentDay - 1) % daysPerPeriod + 1 with daysPerPeriod defaulting to 1. On a default save that value is always 1, so a weekday schedule (five true, two false) ran every day because activeDays[1] was pinned true, and unticking day one stopped the pivot forever, with no error and nothing in the log.

The fix: IrrigationManager:dayOfWeekIndex derives the 1..7 index from the monotonic day modulo 7, never from currentDayInPeriod, so the index actually advances day to day and the weekend-off entries are reachable. The schedule now honours the days the player actually picked.

This is the defect Claude(A) certified on 2026-08-08 at the decompiled base game (Environment.lua:360/381), the second half of the F159/F160 pair. F159 (the skipped-hours charge) is already closed in shipped code; this closes F160.

Verified: 17 new assertions (scs160_schedule_day_index_test.lua): the index advances 1..7 from the monotonic day and ignores the pinned currentDayInPeriod, a weekday schedule rests on Saturday and Sunday, and the mirror case (a schedule active only on Wednesday) runs on Wednesday instead of never. Full suite 465/0, syntax and lint clean. Built and deployed as 1.2.5.71. Not yet observed in a running game.

The TaxMod mirror (its irrigation-expense day index) ships as a companion PR on FS25_TaxMod, so the accrual lands on the same days SCS actually runs a system.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -69,3 +69,8 @@
 - [x] F154 retired the UsedPlus wear bridge and deleted updateSystemWearLevel, leaving system.wearLevel nil on every system. The schedule dialog read that field in updatePerformance and threw on nil during onOpen, leaving the dialog visible with input never initialised (could not be interacted with or closed). Found in-game on the 72h skip test.
 - [x] Fixed by reading wear as system.wearLevel or 0. Per F154 the wear factor is a permanent multiply by one after the retirement, so nil and zero are the same number and no player-observable behaviour changes (F154 invariant 1). F154 bench gains the dialog-consumer assertion. Suite 382/0, syntax clean, built and deployed.
 - [~] Re-open the dialog in-game on the fixed zip to confirm it interacts and closes normally.
+
+## 2026-08-14 (Fred): F160 - the weekly irrigation schedule honours the picked days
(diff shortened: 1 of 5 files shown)


sasha-rf (approved):
Approved. F160, the weekly irrigation schedule now honours the days the player picked.

The day-of-week index was read from env.currentDayInPeriod, which the base game pins at 1 on a default save, so a weekday schedule ran every day and unticking day one stopped the pivot forever, silently. The index is now derived from the monotonic day modulo 7, so the weekend-off entries are actually reachable and the schedule is stable from day one of a fresh save.

The fix is local and honest: a single derivation method used by the schedule check, no change to the schedule model, and the mirror case (a Wednesday-only schedule) is proven to run on Wednesdays rather than never. Bench is 17 new assertions covering the advancing index, the weekday rest, and the mirror case. Full suite 465/0. The TaxMod mirror ships as its companion PR so the expense accrual stays on the same days. Not yet seen in a running game; the fresh-save weekend rest is the owed in-game check.

PR #130, opened 2026-08-14
feat(drilling-window): the rain outlook getter, the sky-reading the advisory wraps

TheCodingDad-TisonK:
The rain outlook getter, the sky-reading the drilling-window advisory wraps.

B1 of the drilling-window build brief: `getRainOutlook(daysAhead)` is a thin published wrapper over the forecast sky-reading SCS already ships. It combines cloud coverage, current rain and the rain-duration heuristic for days 1-2, and the season's rain prior beyond, averaged over the horizon into a single 0..1 likelihood. Always approximate, by law: the result carries `approximate = true` and every consumer must hedge. No new model, no new tracking, and the per-day formula mirrors `getMoistureForecast`'s near-term probability so the two stay in lockstep.

Exposed on the CropStressManager facade for the SoilFertilizer advisory, with a neutral approximate fallback when the weather integration is absent. This is the SCS half of the drilling-window advisory; the SF advisory surface is the companion PR.

Verified: 9 new assertions (scs_rain_outlook_test.lua): the approximate law, wet-outlook-above-clear, the horizon length averaging toward the season prior, the no-cloud-updater degrade, and the wrapper shape. Full suite 474/0, syntax and lint clean. Built and deployed as 1.2.5.72. Not yet observed in a running game.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.71</version>
+    <version>1.2.5.72</version>
 
     <title>
         <en>Seasonal Crop Stress &amp; Irrigation Manager</en>
--- a/src/CropStressManager.lua
+++ b/src/CropStressManager.lua
@@ -796,6 +796,17 @@ function CropStressManager:getYieldKeepFactor(fieldId)
     return keep
 end
 
+-- The drilling-window outlook over the next N in-game days. A THIN published
+-- wrapper over WeatherIntegration's sky-reading (cloud coverage + current rain +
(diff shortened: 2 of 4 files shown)


sasha-rf (approved):
Approved. The rain outlook getter, the sky-reading the drilling-window advisory wraps.

It is exactly what the brief's B1 asks for: a thin published wrapper over the forecast heuristics SCS already ships, combining cloud coverage, current rain and the rain-duration read for days 1-2 with the season prior beyond, averaged over the horizon. The approximate law is carried in the result and every consumer must hedge, which is the honest foundation the whole advisory stands on. No new model, no new tracking, and the per-day formula mirrors getMoistureForecast so the two stay in lockstep.

The facade exposure follows the established wrapper pattern with a neutral fallback. Bench is 9 new assertions covering the approximate law, wet-above-clear, the horizon tail toward the season prior, and the no-cloud-updater degrade. Full suite 474/0. Not yet seen in a running game.

PR #131, opened 2026-08-14
feat(soil-moisture coupling): Arrow-2 compaction + moisture-extreme nutrient uptake, SCS-side

TheCodingDad-TisonK:
The soil-moisture coupling and Arrow-2's compaction half, SCS-side.

Arrow-2 compaction: SoilFertilizerIntegration now reads SF's compaction alongside the existing organicMatter and pH reads and adds a compaction modifier to the critical-moisture threshold, so a compacted field's crop stresses and alerts fire earlier on a drying swing, with low compaction neutral. Same shape as the shipped pH modifier, exactly as the brief's "like the existing evap/stress mods" directs.

The soil-moisture coupling: at drought (below 30% moisture) effective nutrient uptake falls, and a field whose SF nutrient status is Poor is hit harder, scaling the drying-deficit stress accrual in CropStressModifier. All of it is SCS-side reads of SF's getFieldInfo, never a write, so the single-writer firewall the brief certifies holds (SF sole writer of compaction, SCS sole owner of moisture).

Waterlog (> 90%) stays a noted future refinement: the shipped model's contract is "no drying deficit, no stress", and a standalone waterlog stress term would break that contract for every consumer. The drought side is where the deficit model already runs, so the coupling lands cleanly there.

Verified: 8 new assertions (soil_moisture_coupling_test.lua): the compaction bands and cache, the drought availability reduction, the waterlog reduction, the Poor-NPK sharpen, and the SF-absent neutral degrade. The caught-up-hour stress bar was moved out of the drought band so it keeps pinning the elapsed-hour arithmetic. Full suite 480/0, syntax and lint clean. Built and deployed as 1.2.5.73. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -79,3 +79,8 @@
 - [x] The running cost resolved the farm from the local player: on a dedicated server nothing was billed all season (no local player), on a listen server every system was billed to the host (the farmer paid for his neighbours). The bill now resolves each system's owner from its placeable at charge time (getOwnerFarmId), the base-game water-charge pattern. registerIrrigationSystem holds the placeable; deregister removes it when the placeable goes.
 - [x] scs158_dedi_farm_resolution_test.lua at 12 assertions; suite 448/0; deployed 1.2.5.70.
 - [~] In-game (owed): a dedi with pivots on two farms, each farm billed for its own; a sold-and-rebought pivot billed to the new owner.
+
+## 2026-08-14 (Fred): soil-moisture coupling + Arrow-2 compaction (SCS-side)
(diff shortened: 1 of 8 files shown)


sasha-rf (approved):
Approved. The soil-moisture coupling and Arrow-2's compaction half.

The compaction read slots in exactly where the brief says, beside the existing OM and pH reads, and feeds the critical-moisture threshold in the same additive shape, so a compacted field stresses and alerts earlier on a drying swing. The moisture-extreme coupling scales the drying-deficit stress by a nutrient-availability factor that drops at drought and with Poor SF nutrient status. All of it is SCS-side reads of SF's getFieldInfo, never a write, so the single-writer firewall holds.

The waterlog call is the honest part: the brief names it as an extreme, but the shipped model's contract is no drying deficit, no stress, and a standalone waterlog term would break that for every consumer. Leaving it a noted future refinement rather than forcing it in is the right engineering call. Bench is 8 new assertions and the caught-up-hour bar was moved out of the drought band to keep pinning its own arithmetic. Full suite 480/0. Not yet seen in a running game.

PR #132, opened 2026-08-14
feat(SCS-020): transpiration feedback - the growth family's condition scales the transpiration draw

TheCodingDad-TisonK:
SCS-020 transpiration feedback: a field with cells blocked by SF-52's viability mask draws less water and stays wetter, and a field growing at excellent credit draws more and dries faster.

One multiplicative term in the hourly evapotranspiration chain, sourced from SF's published getFieldGrowthSummary (duck-typed, pull-only, neutral 1.0 when SF is absent). The evap loss is split into a soil-evaporation share, which is never scaled, and a transpiration share, which is scaled by the growth condition, so a blocked cell keeps its full soil drying (the pinned invariant from the cold-inversion review). The 0.25..1.25 clamp bounds only the transpiration share.

SCS remains the sole writer of moisture throughout. No new write site, no new persisted state, no player-facing surface, and no new sync. The weights and the share are dials awaiting the spine.

Verified: 10 new assertions (scs020_transpiration_feedback_test.lua): the neutral identity, the formula bands, the share split (soil never scaled), and the SF-absent nil read. Full suite 490/0, syntax and lint clean. Built and deployed as 1.2.5.74. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -84,3 +84,8 @@
 - [x] Arrow-2 compaction half: SF compaction read adds a critical-moisture modifier (compacted fields stress/alert earlier on a drying swing). The soil-moisture coupling: drought (<30%) reduces effective nutrient uptake, Poor SF nutrient status hits harder, scaling the drying-deficit stress accrual. All SCS-side reads of SF getFieldInfo, never a write (firewall holds). Waterlog stays a noted future refinement (would break the no-deficit contract).
 - [x] soil_moisture_coupling_test.lua at 8 assertions; suite 480/0; deployed 1.2.5.73.
 - [~] In-game (owed): a compacted field alerts earlier on a dry spell; a drought on poor soil stresses harder than on good soil.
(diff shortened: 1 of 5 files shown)


sasha-rf (approved):
Approved. Transpiration feedback, the growth family's condition scaling the transpiration draw.

The shape is exactly the brief's: the evap loss is split, only the transpiration share is scaled by the growth summary, and the soil-evaporation share is never touched, so a blocked cell keeps its full soil drying. The duck-typed read of SF's getFieldGrowthSummary is pull-only and neutral when SF is absent, and SCS stays the sole writer of moisture, so the firewall holds.

No new write site, no persistence, no surface, no sync - the member is invisible outside the moisture value it nudges, which is the right size. Bench is 10 new assertions covering the neutral identity, the bands, the share split, and the absent-SF read. Full suite 490/0. Not yet seen in a running game.

PR #137, opened 2026-08-15
fix(irrigator): stop the client log flooding when a hose peer is sold

WizardlyPayload:
Selling a Rainstar while the EMP pump survived filled the client log once a second with
`Unknown entity id` from getWorldTranslation, followed by an arithmetic-on-nil at
phWorldDistance. The hose specialization is pump-only, so the pump never receives a delete
event for its partner and keeps spec.partner, spec.candidate and a registered activatable;
ActivatableObjectsSystem then polls a distance helper into nodes the engine has destroyed.

Two halves, both needed. phResolveHoseJoint now returns a node only if that node exists, and
the liveness gate runs before I3DUtil.indexToObject because that walk reads components and
steps through child nodes itself. Every getWorldTranslation on a hose joint in this file takes
its node from that one resolver, so guarding there covers all of them and the next one added.

invalidateDeadPeer is the other half: forced disconnect, partner and candidate cleared, the
vendored RPC handles cleared, activatable unregistered and hose visual dropped. It runs from
onUpdate, from onUpdateTick before the candidate refresh, and from the top of getIsActivatable,
which is what AOS polls every frame and so notices first. getDistance deliberately does not
clear, since AOS may call it mid-iteration; a dead peer simply stops contributing to the
midpoint. onDelete is now unconditional and idempotent.

Applied onto development's own file rather than copied, so ScsPumpStartActivatable is
untouched.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.76</version>
+    <version>1.2.5.77</version>
 
     <title>
         <en>Seasonal Crop Stress &amp; Irrigation Manager</en>
--- a/src/ScsPumpHoseConnection.lua
+++ b/src/ScsPumpHoseConnection.lua
@@ -46,16 +46,46 @@ local function phIsRainstar(vehicle)
     return string.find(path, "rainstar.xml", 1, true) ~= nil
 end
 
-local function phResolveHoseJoint(vehicle)
+--- Is this vehicle still backed by live engine nodes? A sold vehicle's Lua table outlives its
+--- component entities, so a non-nil partner proves nothing; the component root is the honest
+--- question, because if it is gone every child under it is gone too.
+local function phVehicleAlive(vehicle)
     if vehicle == nil then
+        return false
+    end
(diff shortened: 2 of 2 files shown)


sasha-rf (approved):
## Review: stop the client log flooding when a hose partner dies (SCS #137)

Approved. Verified at the PR head:

- The log-flood root cause is addressed at the source: `phResolveHoseJoint` now gates on `phVehicleAlive` (isDeleted / getIsBeingDeleted / component-root entityExists) before the unsafe `indexToObject` walk, and every getWorldTranslation in the file takes its node from that one guarded resolver.
- `invalidateDeadPeer` clears a sold/destroyed partner or candidate ahead of the range check and the candidate refresh, so a dead partner is not promoted back and the hose prompt goes away the same frame (also called from `getIsActivatable`, which AOS polls).
- `onDelete` is now unconditional + idempotent and calls the new `releaseHoseActivatable`, matching vanilla Dog/AnimalLoadingTrigger cleanup.
- No removed features - `phResolveHoseJoint` is rewritten in place with the guard added. modDesc patch bump to 1.2.5.77.
- Only modDesc.xml + ScsPumpHoseConnection.lua change. No encoding issues.

**Version note:** shares the 1.2.5.77 bump with the heat-sheet PR #138; merge in sequence.

PR #138, opened 2026-08-15
feat(map): Crop Moisture heat sheet - shared ramp, per-level sample step, Soil-parity chrome

WizardlyPayload:
Map to Crop Moisture becomes a sampled heat sheet instead of one dot per field at the field
centroid. A single dot cannot show the thing a player actually wants to see, which is the wet
patch where the irrigator has just been.

One ramp function feeds the tiles, the legend bar and the average-card marker, so the legend
cannot quietly stop describing the map, and one classify helper owns both hairlines so a status
row can never disagree with the card above it. The polygon walk and point-in-polygon test are
ported rather than required across the mod boundary: Crop Stress must keep working with Soil
absent.

Density now moves the sample STEP and not only the point budget. Without that, Low and High
drew more or fewer tiles of the same size and a field edge never got finer whichever was
picked; the step is what the eye reads. Changing it also drops the polygon cache, which is
keyed by step.

Read only: nothing here writes a densmap, and the HUD alert thresholds are untouched.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.76</version>
+    <version>1.2.5.77</version>
 
     <title>
         <en>Seasonal Crop Stress &amp; Irrigation Manager</en>
--- a/src/ui/CsMoistureMapOverlay.lua
+++ b/src/ui/CsMoistureMapOverlay.lua
@@ -1,26 +1,61 @@
 -- =========================================================
 -- CsMoistureMapOverlay
--- Draws per-field moisture centroid dots on the PDA map.
--- Colour-coded by moisture level: green / amber / red.
+-- Map -> Crop Moisture as a Soil-style polygon heat sheet.
+-- =========================================================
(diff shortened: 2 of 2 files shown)


sasha-rf (approved):
## Review: Crop Moisture heat sheet (SCS map overlay, PR #138)

Approved. Verified at the PR head:

- The overlay now fills each owned farmland with a sampled tile grid (point-in-polygon + polygon walk ported into SCS, NOT a cross-mod dependency on SoilFertilizer - verified no Soil require).
- ONE ramp function feeds both the tiles and the legend swatch, so the legend cannot drift from the map. The ramp is display-only: nothing writes a density map and the HUD alert thresholds are untouched.
- Farmland fill uses the field API patterns (polygonPoints, posX/posZ, farmland.id) consistent with the SCS codebase. Sampling step is map-area scaled for consistent density.
- All GuiUtils-normalised positioning. modDesc patch bump 1.2.5.76 to 1.2.5.77.
- Only modDesc.xml + src/ui/CsMoistureMapOverlay.lua change. No encoding issues.

**Version note:** this PR and the hose-log-fix PR #137 both bump SCS to 1.2.5.77. They must be merged in sequence (the second one needs a version bump past the first) - not a blocker, just an ordering note.

PR #139, opened 2026-08-15
Hold Crop Stress field remaps until the player has entered

WizardlyPayload:
## Summary
- Construct the three Crop Stress join packets with a vanilla empty event, so the engine can read them.
- Point the three irrigation vehicle types at the game Vehicle script.
- Hold the hourly field-map walk until the player has entered. Moisture numbers still apply on join.
- Keep the rain-outlook helper that was in the clone and never packed.

## Test plan
- [ ] Dedicated join: no Client.lua nil on Crop Stress join sends
- [ ] Dedicated join: no Crop Stress Vehicle.lua miss
- [ ] Pre-Start: no field remap every 30 seconds
- [ ] Start is a separate test; this PR does not claim Start is fixed

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.77</version>
+    <version>1.2.5.94</version>
 
     <title>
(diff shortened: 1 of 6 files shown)


TheCodingDad-TisonK (approved):
Reviewed: missing emptyNew() on 3 event classes fixed (critical for join), vehicle type filename with dataS prefix corrected, hourly tick and field map rebuild gated on isMissionStarted. All correct.

PR #141, opened 2026-08-16
fix: drop Rainstar control icon names the engine atlas does not have

WizardlyPayload:
## Summary
- Remove `iconName=ICON_DRUM` and `ICON_WWIDTH` from Rainstar.xml. `iconName` is optional.
- Carries Tyson's moisture-cells fix already on development (1.2.5.95). Version 1.2.5.96.

## Test plan
- [ ] Load with Rainstar: no `FS25_SeasonalCropStressICON_` warnings
- [ ] Rainstar axis controls still work

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.95</version>
+    <version>1.2.5.96</version>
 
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):
Approved. `ICON_DRUM` and `ICON_WWIDTH` are not in the engine atlas, so they resolve as `FS25_SeasonalCropStressICON_*` and warn on load. `iconName` is optional on `<controls>`, so removing them is the right fix. Version bump to 1.2.5.96.

PR #144, opened 2026-08-18
fix(scs): complete SCS PDA localization and tab bar spacing

TheCodingDad-TisonK:
## What

The Crop Stress PDA (Esc RF page and the legacy Overview / Irrigation page) ships with localization gaps and a cramped tab bar. This fills the gaps and matches the tab bar to the sibling Soil and Market pages.

## Localization

- English gained the only two `cs_rf_pda_*` keys the CS desk used through fallbacks: `cs_rf_pda_crop_mixed` and `cs_rf_pda_pivot_switch`.
- The hardcoded `CROP CONSULTANT` link text on the legacy PDA now reads from `cs_pda_btn_consultant`, which exists in all 26 languages.
- Every `cs_pda_*` and `cs_rf_pda_*` key now exists in every one of the 26 translation files. The 17 languages that were missing the whole Esc `cs_rf_pda_*` block (75 keys) got it filled with `[EN]` placeholder values, the existing repo convention, so no PDA string can render as a raw key in any language.
- Seven translation files that carried a UTF-8 BOM are now BOM-free, matching the rest of the tree.

## Spacing

- `CsPDAScreen.xml` tab bar now uses the Soil / Market geometry: 58px bar at y=668, divider at y=666, labels cleared off the bar edge, content panels at 668 tall. The tab labels and the footer buttons stop crowding each other.

## Verified

- Built and deployed. All 26 translation files parse as well-formed XML, BOM-free, with identical `cs_pda_*` / `cs_rf_pda_*` key coverage.
- Lua test suite: 499 assertions passed.
- Not yet tested in game; log check and a visual pass are still owed.


sasha-rf (approved):
Approved. Verified: every cs_pda_* and cs_rf_pda_* key now exists in all 26 translation files (XML well-formed, BOM-free), the hardcoded CROP CONSULTANT link reads cs_pda_btn_consultant, and the CsPDAScreen tab bar matches the Soil/Market geometry. Builds, 499 test assertions pass.

PR #145, opened 2026-08-18
feat(scs): render moisture map through the engine DMV overlay

TheCodingDad-TisonK:
## What

The map's Crop Moisture page still drew a grid of sampled tiles. Soil moved to the engine density-map visualization (DMV) overlay; this does the same for SCS.

## How

- `CsMoistureMapOverlay` now creates the same double-buffered DMV overlay pair Soil's PDA uses (`createDensityMapVisualizationOverlay`, async build, buffer swap on ready).
- A dedicated display bitvector (`CS_PDA_MoistureDisplay`) is band-painted per field polygon with each farmland's moisture scalar, and the DMV renders it over the map rect as a crisp per-pixel overlay.
- The band map is separate from `CropStressValueMap` on purpose: the sim map encodes moisture linearly over 254 raw steps, where the driest pixels land in DMV state 0, the off-field transparent band. Band-painting a dedicated map floors every field at band 1, so a bone-dry field stays visible, and it never touches the persisted sim data or the save format.
- The DMV path computes the sidebar band counts itself (they used to come from the tile sampler), so the legend stays live.
- The sampled tile grid is retained as the engine-absent fallback.

## Verified

- Built and deployed. Lua 5.1 syntax clean, 499 test assertions passed.
- Not yet tested in game; log check and a visual pass are still owed.


sasha-rf (approved):
Approved. The moisture map renders through the engine density-map visualization overlay (Soil pattern): double-buffered overlays, a band-painted display bitvector separate from the persisted sim map, tile fallback kept, sidebar band counts computed from the DMV pass. Lua 5.1 syntax clean, 499 test assertions pass.

PR #146, opened 2026-08-18
fix(scs): map sidebar l10n strings and block spacing

TheCodingDad-TisonK:
## What

The map sidebar's legend, average card and info box still drew missing l10n strings, and the three blocks sat too close together. This fixes both.

## Localization

- 16 `cs_map_*` keys used by the sidebar draw calls existed in no translation file: `cs_map_status_dry/good/wet`, `cs_map_density`, `cs_map_density_low/med/high`, `cs_map_legend_dry/good/wet`, `cs_map_avg_title`, `cs_map_avg_ok`, `cs_map_avg_irrigate`, `cs_map_avg_toowet`, `cs_map_info_title`, `cs_map_info_line`. All 16 now exist in all 26 language files; non-English files carry `[EN]` placeholder values per the repo convention.
- The overlay's `tr` helper now rejects unresolved keys the engine can hand back (the raw key, the `$l10n_` literal, and "MISSING ..." variants), so a missing key always falls back to English and never renders as a raw string.

## Spacing

- The legend, average card and info box are now separated by a dedicated 18px card gap instead of the previous 4-8px, so the sidebar reads as separate blocks.

## Verified

- Built. All 26 translation files parse as well-formed XML, BOM-free, with the new keys present. Lua 5.1 syntax clean, 499 test assertions passed.
- Deploy to the active mods folder is blocked while the game is running (the zip is locked); the build zip is ready. Not yet tested in game.


sasha-rf (approved):
Approved. Verified: the 16 cs_map_* keys the sidebar draw calls use now exist in all 26 language files (XML well-formed, BOM-free), the tr helper rejects unresolved keys so a missing key falls back to English, and the legend / average card / info box are separated by the 18px card gap. Lua 5.1 syntax clean, 499 test assertions pass.

PR #147, opened 2026-08-18
fix(scs): drop the bogus WOOD_CHIPPER_IDLE sound from the pump turn-on

TheCodingDad-TisonK:
The irrigator kit's Aggregat.xml referenced sound template WOOD_CHIPPER_IDLE in turnOnVehicle.sounds.work, which the game does not have, so the kit errored at load (found in the RF Dev server log: 'Sound template WOOD_CHIPPER_IDLE was not found'). The kit's own pump idle plays through pumpIdleSound; the turn-on block now carries no sound. One line removed.


sasha-rf (approved):
Approved. Verified: `Aggregat.xml` referenced sound template `WOOD_CHIPPER_IDLE`, which the game does not have (load-time error in the RF Dev server log); the reference is removed and the kit's own pump idle still plays through `pumpIdleSound`. XML well-formed.

PR #154, opened 2026-08-24
docs: correct RealisticWeather integration description

TheCodingDad-TisonK:
## Summary
- README table claimed SCS sourced moisture from RW's density map and wrote irrigation back to RW
- Since the SCS-018 unwind, SCS owns its own moisture simulation and only reads temperature/rainfall from RW
- This outdated line was the likely source of player confusion in nitro's report about "SCS moisture does not follow the weather"

One-line docs fix, no code change.


sasha-rf (approved):

PR #155, opened 2026-08-24
fix: strip mod prefix from Rainstar vehicle type checks

TheCodingDad-TisonK:
## Summary

- FS25's TypeManager prefixes runtime `vehicle.typeName` with `ModName.` (e.g. `FS25_SeasonalCropStress.rwsmPlayRainstar`), but all Rainstar code paths compared against the bare name
- Added `matchesType()` helper that strips the prefix before lookup
- Fixed all four affected files: IrrigatorSectorIntegration (sector tick), SprayerIntegration (exclusion guard), ScsPumpHoseConnection (hose partner check), ScsRainstarPumpPower (pump diagnostics)

## Root cause

`TypeManager.lua:259` returns `modName .. "." .. typeName`, so at runtime the Rainstar's type is `FS25_SeasonalCropStress.rwsmPlayRainstar`. Direct table lookups against the bare name `rwsmPlayRainstar` always returned nil, silently disabling the irrigator sector path and the sprayer exclusion guard.

## Test plan

- [x] Deployed and tested in-game: irrigator sector fires every 500ms, log confirms `[CropStress] Irrigator sector` lines with 24 cells per tick
- [x] Pump power confirmed working standalone (pump connected + started, no tractor required)
- [x] Lua 5.1 syntax check passes (pre-commit hook)

### DIFF
--- a/src/IrrigatorSectorIntegration.lua
+++ b/src/IrrigatorSectorIntegration.lua
@@ -24,6 +24,15 @@ IrrigatorSectorIntegration.VEHICLE_TYPES = {
     rwsmPlayRainstar = true,
 }
 
+--- Match a vehicle's typeName against VEHICLE_TYPES, stripping the mod prefix
+--- FS25 uses "ModName.typeName" at runtime (TypeManager.lua:259).
+function IrrigatorSectorIntegration.matchesType(typeName)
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):

PR #157, opened 2026-08-25
fix: crop moisture map shows per-pixel moisture, irrigator water, and seeds at load

TheCodingDad-TisonK:
## What

The Crop Moisture map did not render irrigator water, opened completely blank on a fresh save, and showed each field as one flat average instead of per-pixel moisture.

## Commits

- fix: irrigator water accumulates past the value-map raw-step floor
- fix: seed the moisture map from the store at load so it is not blank
- fix: restore quantised irrigator accumulation on the map write path
- fix: seed per-pixel relief variation so the moisture map is not flat averages

## Why

Three defects in the same surface:

1. `applyWaterAtCell` wrote each tick's gain straight into the 2m value map. One irrigator tick's gain, spread over the whole sector area, is about 200 times smaller than a single raw step on the map (~0.0039 moisture), so the write re-encoded the same raw value and the map never moved while the cell store kept rising.
2. A fresh value map has no pixels until the first write. The overlay opened empty, and the first irrigator write called `migrateFieldToMap`, which base-coats the entire polygon at its field aggregate before the wet pixels land.
3. The base coat left every field one flat average. Relief variation was never painted onto the map.

## How

- `applyWaterAtCell` accumulates the sub-step remainder per map pixel and spends only whole raw steps through `CropStressValueMap.quantiseDelta`, the same shape the hourly weather path already uses with `mapPending`.
- `seedMapFromStore` paints every tracked field onto the map after the store scalars are restored. A map restored from its own `.grle` is skipped so per-pixel truth is never flattened.
- `_seedMapRelief` paints per-pixel relief variation over the base coat at load: low ground reads wetter by the same SENS/MAX the store's relief pass uses, and the offsets sum to about zero so the field mean is unchanged.

## Verified

Built, deployed, Lua 5.1 syntax clean (pre-commit). In-game: overlay starts fully painted with within-field variation; wet patch appears under an active irrigator.

### DIFF
--- a/src/CropStressManager.lua
+++ b/src/CropStressManager.lua
@@ -742,6 +742,13 @@ function CropStressManager:loadFromXMLFile()
     else
         self.saveLoad:loadFromXMLFile()
     end
+
+    -- SCS-039: now that the store scalars are restored, paint a fresh map from
+    -- them so the moisture overlay is not blank until the first write. A map
+    -- restored from its own .grle is skipped: it is already the per-pixel truth.
+    if self.soilSystem ~= nil and self.soilSystem.seedMapFromStore ~= nil then
+        self.soilSystem:seedMapFromStore()
+    end
 end
 
 -- ============================================================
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):

PR #158, opened 2026-08-25
feat: PDA and RF Esc desk report and control Rainstar irrigation

TheCodingDad-TisonK:
## What

The SCS irrigation surfaces did not report a field as irrigated when it was being watered by a parked Rainstar with the Aggregat pump, and the RF Esc PIVOT card rendered an empty seat for such a field. Both the dedicated SCS PDA and the RF Esc Crop Stress desk only knew about placed pivots and drip lines.

## Commits

- fix: PDA reports fields actively watered by a Rainstar as irrigated
- fix: RF Esc Crop Stress desk also reports Rainstar-watered fields
- feat: RF Esc PIVOT card shows the Rainstar and can stop it

## Why

The covered-fields set came only from `IrrigationManager.systems`, which registers placed pivots and drip lines. A Rainstar is a vehicle irrigator driven by `IrrigatorSectorIntegration`; it watered the field but never reported it, so the irrigation tab showed "--", the row said no irrigation system was connected, and the PIVOT card (a Reinke placeable controller) had no seat to show.

## How

`IrrigatorSectorIntegration` records every farmland it waters each tick and which vehicles watered it, exposed through `getActiveWateredFields` and `stopForField`. Both `CsPDAScreen` and the RF Esc guest merge that set into their irrigated flag and counts. When no placed pivot covers the selected field but a Rainstar is actively watering it, the PIVOT card shows a Rainstar seat (title, watering status, coverage) and enables the Stop remote, which turns the active Rainstar off. New l10n keys ship in all 26 languages.

## Verified

Built, deployed, Lua 5.1 syntax clean (pre-commit). In-game: a field under an active Rainstar reads as irrigated in both surfaces, and the PIVOT card shows the Rainstar seat with a working Stop.

### DIFF
--- a/src/IrrigatorSectorIntegration.lua
+++ b/src/IrrigatorSectorIntegration.lua
@@ -46,6 +46,10 @@ function IrrigatorSectorIntegration.new(manager)
     self.isInitialized = false
     self._accMs = 0
     self._waterFillType = nil
+    -- farmlandId -> true for every field this tick actually watered. The PDA
+    -- reads this so a parked Rainstar shows its field as irrigated the same way
+    -- a placed pivot or drip line does; cleared at the top of each tick window.
+    self._activeWateredFields = {}
     return self
 end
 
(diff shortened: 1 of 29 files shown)


sasha-rf (approved):

PR #160, opened 2026-08-26
fix(scs): probe every surface for the RealisticWeather handle

TheCodingDad-TisonK:
Follow-up to the Discord testing wave report (nitro + J. Duke FF1-8505): SCS moisture does not follow RealisticWeather.

The RW detection in `WeatherIntegration:detectOptionalMods` probed only `getfenv(0)["g_realisticWeather"]` - the per-mod-scoped global read this repo's own CLAUDE.md documents as unreliable cross-mod (proven with the NPCFavor integration). If that read is invisible, SCS never arms `realisticWeatherActive`, falls back to the vanilla environment, and moisture never follows RW - exactly the report.

**Fix:** a single `_rwHandle()` resolver probes every surface the handle could live on (the bare global, the `g_currentMission` bridge, `getfenv(0)`, and the `g_weatherSystem` family the same way), records which matched on `_rwSurface`, and the detection plus all three weather reads (temperature, humidity, rain) use the same resolver so they cannot disagree. A csDebug session will now log which surface actually resolved, confirming or refuting the cross-mod path.

New `weather_rw_detection_spec_test.lua` at 12 assertions (nil case, bare global, mission bridge, g_weatherSystem fallback, and read-path agreement). Suite: 465 passed, 0 failed, plus one pre-existing load error in `scs039_delegation_test.lua` (its own `worldToPixel` stub gap, unrelated to this change and present on development). Deployed.

The decisive confirmation is still the in-game check: load with RealisticWeather, run `csDebug`, and read the weather line - it should now show "RealisticWeather detected (g_realisticWeather)" and moisture should track RW's rain. Not yet verified in a game session.

### DIFF
--- a/src/WeatherIntegration.lua
+++ b/src/WeatherIntegration.lua
@@ -121,19 +121,49 @@ end
 -- OPTIONAL MOD DETECTION
 -- ============================================================
 function WeatherIntegration:detectOptionalMods()
-    -- FS25_RealisticWeather detection
-    -- This mod exposes g_realisticWeather global with enhanced weather data
-    if getfenv(0)["g_realisticWeather"] ~= nil then
+    -- FS25_RealisticWeather detection. The mod publishes the bare global
+    -- g_realisticWeather (RealisticWeather.lua:109); whether that bare global is
+    -- visible here depends on the engine's mod-environment scoping, so probe every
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):

PR #161, opened 2026-08-27
Add moisture HUD hide/show toggle to the Control Center

TheCodingDad-TisonK:
Reverses the earlier 'no Control Center button' decision for the moisture HUD. The moisture HUD is hidden by default and its physical key is suppressed under MasterHUD, which left MasterHUD players unable to open it at all. This adds a CS_TOGGLE_HUD hide/show delegate (via g_cropStressManager:onToggleHUD, live caption) to restore per-mod access from the Control Center. __rfHudAct still gates the physical key; only the 'no button' half is reversed.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -504,10 +504,12 @@ addModEventListener({
 -- ---------------------------------------------------------
 -- Realistic Farming Control Center: publish runnable delegates.
 --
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #163, opened 2026-08-31
feat: vendor Reinke A22 pivot into SCS, swap shop pivot to water tank/pump

TheCodingDad-TisonK:
Vendor the Antler22 Reinke A22 center-pivot into SCS and swap the shop pivot.

What changed:
- The Reinke A22 pivot asset (3/5/10-tower variants, ~115/188/369m radius) now ships inside SCS at `placeables/reinkeA22/`, replacing the placeholder `centerPivot` in the shop.
- New `reinkeIrrigationPivot` placeable type + specialization registered in modDesc; the pivot self-integrates with `IrrigationManager` at load (its lua calls `registerIrrigationSystem`), so the Esc PIVOT card and SCS scheduling work against the same spec table.
- The 12 `REINKE_*` actions and their KP default bindings are declared; `reinke_pivot_*` and `input_REINKE_*` strings ship in all 26 translation files.
- The water pump now wears the base-game `waterTankLevel02` model (via `$data`), so the pump is a water tank.
- Removed the old placeholder `centerPivot` and the `dripIrrigationLine` placeable. The shop now carries two placeables (water tank/pump and pivot) plus the three irrigator vehicles (Rainstar, Aggregat, pull bar), which are untouched.
- Esc PIVOT dial-face decal now resolves from SCS's own directory; UsedEquipmentMarketplace references updated to the Reinke type.

Not yet tested in game. Lua syntax and the test suite pass (the one failing scs039 case is a pre-existing harness gap, unrelated to this change).


sasha-rf (approved):

PR #166, opened 2026-09-01
fix: Esc pivot Fit, Door, invert chips, and captions

WizardlyPayload:
## Summary
- Esc Fit on an owned Reinke works in playtest: the pivot card hit-tests Fit so the click cannot fall through to Door.
- Reinke registers toggleDoor (menu, no standing at the box) plus sweep min/max helpers. Dial copy is ASCII (Arm / Sweep start / Sweep end).
- Rain-key command and result events use Event.new so the engine can run them. Server apply still goes through applyRainKeyCommand. Finite-water engine from development is unchanged.
- Door, Power, Spray, End gun, Start, and Stop invert the live chip colours when that function is on. Locked chips stay grey.
- Sweep start sits above the min pair, Sweep end above the max pair, and Rain key sits on Fit, in the same type as the pivot status lines.

Based on origin/development (finite water + rain-key engine). Playtest Crop Stress zip is 1.2.5.104. This branch is 1.2.5.104 on current development.

If another mod creates the Esc door first, that host page also needs the Fit chips and caption texts. Crop Stress guest paints invert and fills the captions once those ids exist.

## Test plan
- [ ] Full start (not only a menu reload).
- [ ] Esc Crop Stress, owned Reinke: Fit changes to Remove and Trip unlocks.
- [ ] Door, Power, Spray, End gun, Start, Stop invert when that function is on.
- [ ] Sweep start / Sweep end / Rain key captions are readable and do not cover the title.
- [ ] Dedicated plus client: Fit still applies on the server (label and tablet rain-key line agree).

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.97</version>
+    <version>1.2.5.104</version>
 
     <title>
         <en>Seasonal Crop Stress &amp; Irrigation Manager</en>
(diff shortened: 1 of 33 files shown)


sasha-rf (approved):

PR #167, opened 2026-09-02
fix: water source names and pump register so saves load

WizardlyPayload:
## Summary
- Irrigation Suite WATER SOURCES rows used `getText` on a key that does not exist, so the game printed `Missing 'cs_water_sou...`. The getter now asks `hasText` for the existing Water Source string, then `getText`, else the English words.
- Water pumps called `registerWithIrrigationManager` as an instance method that was never registered on the placeable type, so load hung after the last vehicle. The function is now registered and called on the class.

Playtest pack was Crop Stress 1.2.5.108. Farm Tablet is unchanged.

## Test plan
- [ ] Full start (zip load, not a workshop-file reload).
- [ ] Save loads past the last vehicle; both pumps register as finite water sources.
- [ ] Farm Tablet Irrigation Suite, Operations: WATER SOURCES rows say Water Source (or the localized string), not Missing. Hours and connected count unchanged.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <modDesc descVersion="105">
     <author>TisonK</author>
-    <version>1.2.5.104</version>
+    <version>1.2.5.108</version>
 
     <title>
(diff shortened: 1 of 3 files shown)


sasha-rf (approved):

PR #170, opened 2026-09-04
feat(SCS-037): caught-up hour — refuse lower/equal/invalid hour keys

TheCodingDad-TisonK:
## SCS-037 — lower-hour admission (build brief amendment)

Travelled irrigation-set card (`d34a9dd`), Stage 6B certified. Host-side, one of four.

### What changed
- **`src/CropStressManager.lua`**
  - `elapsedHoursFrom` now returns **0** for a lower, equal, fractional-under-one or invalid observed key; **1** for first observation (nil/nonnumeric/negative prior); positive floored span capped at 168 otherwise.
  - The `update` hour branch advances `lastHourKey` **only on a positive span** (no frontier rewind) and calls `onHourlyTick` only then. **No function-level return** — the HUD/settings tail still runs on refused frames.
- **`tools/test/lua/scs037_caught_up_hour_test.lua`** — elapsed-count contract updated: `backwardsIsZero`/`sameKeyIsZero` replace the old is-one cases, plus invalid-observed and fractional-under-one assertions. Downstream span assertions unchanged.

### Why
A developer `gsTimeSet` to an earlier hour of the same monotonic day was replaying already-consumed hourly work. This makes the detector positive-only; `onHourlyTick`'s existing clamp (which turns 0 back into 1) is never reached on a refusal.

### Verification
- `✓ scs037_caught_up_hour_test.lua (45 passed, 0 failed)` (42 → 45).
- Full suite: 0 assertions failed. The single load error (`scs039_delegation_test.lua`, `worldToPixel` nil) is pre-existing on `development` — unbuilt SCS-039 surface, out of scope for this PR.

### Boundaries (per brief)
No new persistence, sync event, setting, authority, grid or player surface. Live-session scoped. First observation and positive forward behaviour unchanged. Release remains locked under the irrigation-set gate.

### DIFF
--- a/src/CropStressManager.lua
+++ b/src/CropStressManager.lua
@@ -124,21 +124,23 @@ CropStressManager.MAX_CATCHUP_HOURS = 168
 
 --- Hours elapsed between two hour keys, as the hourly path should charge them.
 --- Pure and static so the bench can drive it without a mission.
---- Returns 1 (today's behaviour exactly) for the first tick, for a key that did
---- not advance, and for any key that moved backwards.
+--- Returns 1 for the FIRST observation (nil, nonnumeric or negative prior key).
+--- Returns 0 — no hourly work — for a lower, equal, fractional-under-one or
+--- invalid observed key: the live clock never re-runs already-consumed hours.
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):

PR #171, opened 2026-09-04
Pivot Auto/Manual, apply-on-load, Pro Staff twenty-card door shell

WizardlyPayload:
## Summary
- Apply the saved watering window as soon as the player is in the world, not only on the next hour tick.
- Persist Auto/Manual. Manual skips both auto start and auto stop. Esc chip plus schedule Save sync.
- Realistic Farming Esc tab sits after Map.
- Carry the Pro Staff twenty-card shell on this door copy.
- Built on current development, including the skipped-hour clock rule already merged there. This PR does not rewind that rule.

## Test plan
- [ ] Full client start.
- [ ] Load a save inside a covering window: water runs without waiting for the next hour.
- [ ] Auto/Manual chip behaves like the real knob. Manual does not auto start or auto stop.
- [ ] Saving a covering window applies now.
- [ ] Skipped / rewind hours still do not replay (development skipped-hour merge stays in place).
- [ ] Pro Staff page shows twenty cards when Crop Stress hosts.


sasha-rf (approved):

PR #181, opened 2026-09-06
Group Crop Stress field lists by GPS outline and match the Soil Esc door

WizardlyPayload:
## Summary
- Crop Stress owned-field lists (Esc table, tablet overview, consultant, HUD) use Soil's GPS-outline groups when Soil is present.
- The shared Realistic Farming Esc door copy matches the working Soil host so leftover chrome from the last page cannot stick when Crop Stress wins first-writer.
- HUD reads cached groups only and does not start outline walks on the draw tick.

## Test plan
- [ ] Full start.
- [ ] With Soil loaded, Crop Stress field lists show the same joined outline as Soil (Montana 86-87).
- [ ] Without Soil, lists stay per farmland.
- [ ] Esc Realistic Farming: switch Crop Stress / Market / Worker Costs; leftover chrome from the previous page must hide.
- [ ] Crop Stress Help footer still appears on that module.

### DIFF
--- a/gui/CropConsultantDialog.lua
+++ b/gui/CropConsultantDialog.lua
@@ -134,6 +134,21 @@ end
 -- Populates fieldListContainer with up to 5 fields sorted by risk.
 -- ============================================================
 
+--- BUILD 19:23 (George CLOSED DESIGN 19:12): the Soil list-row merge module on the mission bridge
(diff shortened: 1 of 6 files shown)


TheCodingDad-TisonK (approved):
Sasha review: field grouping via Soil merge bridge matches George's closed design, nil-safe when Soil absent. Approved.

PR #185, opened 2026-09-11
feat: farm-scoped irrigation snapshot carries rain-key + running/paused

TheCodingDad-TisonK:
## What (Wizard/Ash inbox item: irrigation host remainder)

The Farm Tablet Irrigation Suite is Wizard-complete; this is the **Crop Stress host** remainder. The farm-scoped snapshot the tablet reads was dropping rain-key + running/paused, forcing the tablet to peek a second public list. This closes that gap.

### Part 1 - farm-only snapshot carries the fields
- **`copyIrrigationSystemRow` is now the single row builder.** The legacy no-arg public copy (`CropStressManager:getIrrigationSystems()`) now routes through it too, so the two copies can never drift again - that drift *is* this bug.
- The builder emits **rain-key readout** (`rainKeyFitted/State/Tripped/TripMm/AccumulatedMm/DryElapsedMinutes`, `weatherReadable`) and **composed `activityState`** (`RUNNING` / `RAIN_PAUSED` / `OFF`) + `pauseReason` / `nextWakeKind` / `stateRevision` **unconditionally**. Only `ownerFarmId` / `waterSourceId` / `stopReason` remain private.
- **`CropStressIrrigationStateEvent`** write/read extended in **matching order** so pure clients receive these fields in their private snapshot (previously they didn't cross the wire at all).

### Part 2 - same truth for pumps / remaining water / schedule surfaces
Verified rather than re-plumbed: the host PDA/schedule guest (`CsRfPdaGuest`) reads the **same live `irr.systems`** objects, and pumps/remaining water come from the **shared `getIrrigationWaterSources` getter**. So once Part 1 makes the snapshot carry the same rain-key/activity truth, the tablet and the host surfaces agree by construction. The guest's "Watering now: On/Off" label + rain-key warning line is left unchanged (Wizard-owned layout, `[BUILD 22:43]`); reversing that deliberate simplification would touch the Esc-door PDA subsystem and is flagged as an optional follow-up, not done here.

## Guardrails honored
Fast-path correction of an already-decided data contract (not new design). No FarmTablet rebuild requested. SF-54 / growth screen untouched. No new hydrology surface.

## Verify
All FS25 stream/API calls read from source first; MP write/read order matched (silent-corruption rule). Lua 5.1 gate passed; built and deployed. Needs in-game MP verification (client sees running/paused + rain-key on the tablet without the public-list peek).

### DIFF
--- a/src/CropStressManager.lua
+++ b/src/CropStressManager.lua
@@ -1063,57 +1063,13 @@ function CropStressManager:getIrrigationSystems(farmId)
         end
         return nil
     end
+    -- Legacy no-arg public copy: route through the single row builder so the
+    -- public list and the farm-scoped private snapshot can never drift again.
+    -- includePrivate = false keeps the all-public field shape older FarmTablet
+    -- versions expect (no ownerFarmId / waterSourceId / stopReason).
     local out = {}
     for _, sys in pairs(irrMgr.systems) do
-        local covered = {}
-        if sys.coveredFields ~= nil then
-            for i = 1, #sys.coveredFields do covered[i] = sys.coveredFields[i] end
-        end
-        local schedule = nil
-        if sys.schedule ~= nil then
-            local days = {}
-            if sys.schedule.activeDays ~= nil then
-                for i = 1, #sys.schedule.activeDays do days[i] = sys.schedule.activeDays[i] end
-            end
-            schedule = {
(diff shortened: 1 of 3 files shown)


sasha-rf (approved):

UTF-16 scan: 3 file(s) read at the PR head, all clean.

PR #187, opened 2026-09-13
fix(SCS-023): Irrigate Now checks the real rain-key state revision

TheCodingDad-TisonK:
## What

`IrrigationManager:applyIrrigateNowTransaction` read `system.StateRevision`, a field nothing ever writes. Every writer (fit, remove, trip, dry reset) and the published snapshot row use `rainKeyStateRevision`. So the fitted-pivot `stale_confirmation` gate could never match a live revision, and every Irrigate Now result reported `stateRevision 0`. The bench masked it by setting the phantom field.

Both reads (`IrrigationManager.lua:906`, `:933`) now use `rainKeyStateRevision`.

## Evidence

- `scs023_irrigate_now_chain_test.lua` sets the real field and now also asserts the transaction result and the snapshot row report the same revision. Control run on the pre-fix source: 3 fail.
- Suite 1923 passed / 0 failed. The load error in `facade_readapi_test.lua` (missing `copyIrrigationSystemRow` stub) is pre-existing on `development` and untouched here.

## Owed in game

A fitted pivot's Irrigate Now confirm with a real revision, and a stale one refused, over the real events.

H1 irrigation unit, existing approved work. No version bump.


sasha-rf (approved):

UTF-16 scan: 2 file(s) read at the PR head, all clean.

PR #188, opened 2026-09-13
feat(SCS-039): save view cut and one restore barrier for the provider envelope

TheCodingDad-TisonK:
## What

SCS-039's own on-disk carrier and restore barrier, per the 2026-09-09 Design return section 1 (steps 1 to 5). The engine-free capture/commit/selector core existed; this is its live file and reload integration plus SCS-041 leaf admission.

- `CropStressManager:ensureMissionWaterSaveCut(reason)`: the one cached immutable save view. Clean view reused as the same object by the pump save (`placeables/waterPump/waterPump.lua`), StateLedger (`buildStateTable`) and career XML (`saveToXMLFile`). Dirty view: settle fitted groups once, refresh aggregates, pack both pending stores and the absorption leaf, write the generation-qualified native image through the real `saveNativeMap` receipt (plus an on-disk existence check), commit. Failure advances no generation, records PENDING_ONLY, leaves the cut dirty. The cut is dirtied once per simulation frame; a save act is synchronous within a frame so its callbacks share one view. No token, timer or nonce.
- Own XML persists the full logical view under `careerSavegame.cropStress.moisture` (current and previous COMPLETE envelopes, PENDING_ONLY row, absorption leaf with its 10 keys and digest) and keeps every legacy key and field row. The ledger table carries the identical view (`moistureEnvelope`). Outer schema 3, leaf schema 2, legacy scalar schema 2 migrates with no absorption allowance.
- `loadFromXMLFile` / `applyStateTable` stage snapshots instead of publishing water state. HUD, schedules and the NPC row still apply on arrival. `collectMoistureCandidates` attaches full payloads and rebuilds each digest; a mismatch rejects that candidate. `selectMoistureCarrier` now has its real caller (`restoreMissionWater`), which probes only the selected generation's native file (`CropStressValueMap:loadGenerationFile`: exists, engine load returns true, width matches `mapWidth`); a newest valid compact without a usable image degrades on its own pending state and declines the live map rather than pairing with another generation's file.
- ONE barrier, `CropStressManager:tryRestoreMissionWater`, gated on `compactReady`, `fieldsReady`, `mapReady`, `settingsLoaded` (new `markSettingsLoaded` from the onStartMission hook) and the absorption freeze. Applies the selected envelope plus any identity-matching PENDING_ONLY overlay, hands the leaf to `loadAbsorptionWindow` with the enclosing provider mode and grain, current hour key and live field membership, prunes missing fields, seeds a fresh map once, then initializes SprayerIntegration and IrrigatorSectorIntegration (moved out of `initialize`). Hourly tick, schedule apply-on-load, the Irrigate Now event and the host wrapper refuse work until `isMissionWaterReady()`. An apply with no enumerated field is not counted.
- `SoilMoistureSystem:adoptNativeGeneration` / `declineNativeCarrier` are the load-time owners; the in-mission fail-closed path is untouched. PR #187's two lines in `applyIrrigateNowTransaction` are untouched (the gate sits in the event `run` and `runIrrigateNowHost`).

## Evidence

- `tools/test/lua/scs039_save_view_restore_test.lua`, 187 assertions bound to the real owners (fake engine handle and a spy on the native receipt only): cut reuse and recapture, failed native write (generation held, PENDING_ONLY, dirty, recovery), XML and ledger full-view round trips through the manager barrier, native unavailable degrade with own pending and map decline, corrupt digest falls to the previous pair, conflicting COMPLETE payloads invalidate the generation, PENDING_ONLY overlay applied only on exact identity, leaf admitted only against the matching provider identity (mismatch stands down, expired window), legacy scalar migration with no allowance, barrier once and only when all five facts are set in four arrival orders, no-field not counted, compact staged once.
- Control: with `src`, `main.lua` and `placeables` stashed, the new bench errors at its first call (`ensureMissionWaterSaveCut` nil) and the two adapted persistence benches error on the missing restore step. Suite on old code 1930/0 with 3 files erroring.
- Suite on this branch: 2154 assertions, 0 failed, 54 files. Syntax check 54 files OK. `facade_readapi_test.lua`, which failed to load on `development` (missing `copyIrrigationSystemRow` stub), now loads and passes; the fix is a stub in that bench's fake irrigation manager.
- API: `loadBitVectorMapFromFile` / `getBitVectorMapSize` / `saveBitVectorMapToFile` / `fileExists` verified against `dataS/scripts_decompiled` (`densityMaps/InfoLayer.lua`, `field/FieldGroundSystem.lua`, `debug/WrapFunctions.lua`); `XMLFile:setString/getString/setBool/getBool` in `xml/XMLFile.lua`; `isMissionStarted` is set in `BaseMission:onStartMission` before our appended hook, so the field updater always runs the frame after compact staging.

## Owed in game

- Real career save and reload (SP) with the map live: generation files `csMoistureMap.gN.grle` written and adopted, envelope round trip in `careerSavegame.xml`, `[CropStress]` restore line reports TRUTH with the expected generation.
- Save interruption behaviour, added disk size and save time. Unreferenced generation files are NOT cleaned yet (the SDS 3.7 cleanup clause is outside this item), so disk grows one native image per save until that lands.
- StateLedger present: both mirrors identical, ledger and XML candidates dedupe.
- Dedicated server and MP client join after the barrier.
- Legacy save (pre-envelope) import as generation 0.


sasha-rf (approved):

UTF-16 scan: 15 file(s) read at the PR head, all clean.

PR #190, opened 2026-09-17
feat(RSF-F244): moisture generation cleanup: no-open probe, three slot files, exact-true ledger registration

TheCodingDad-TisonK:
## What this is

The certified RSF-F244 handoff (SCS-039 section 3.7 cleanup, plus F250 item 2), built on PR188's save/view/restore barrier. Source of intent: design wave F191/F244 (ledger e38cda7) and its `DELIVERY.md` and `CURRENT-PICKUP-STATUS.md`. Bob's intake is FAST TRACK.

**Everything native here sits behind the release gate.** `cs_grid_concordance` is off by default, so ordinary saves run the cell store and none of the native changes reach them. The StateLedger fix is the exception: it applies to everyone.

## The three defects

1. **A stale moisture picture could come back as current truth.**
   - `CropStressValueMap:initialize` opened `csMoistureMap.grle` at field-ready, before the restore barrier had read any envelope.
   - On the barrier's no-candidate branch that legacy image survived, even on a save that had already moved to generation storage.
2. **Every save left the previous moisture image on disk for good.**
   - That is one map-sized file per save: about 1 MB on a 2 km map, 16 MB on 8 km.
   - The mod cannot delete them. The mod environment wraps `deleteFile` so it only reaches `modSettings/<modName>` (decompiled `mods.lua:708-733`, `protectedDelete`).
3. **A refused StateLedger registration was reported as registered.** The bridge ignored `registerModule`'s boolean return.

### 1. The probe opens no file
- `initialize()` always stands up a fresh map. It never reads `LAYER_DEF.file` and leaves `loadedFromSave` false.
- **One loader:** `loadNativeFile(dir, filename, expectedWidth)`, called only from inside the barrier. It requires, in order:
  - a whole-string accepted name: exactly `csMoistureMap.grle`, a positive `csMoistureMap.g<N>.grle` with no leading zero (so `g0` and `g01` are refused), or `csMoistureMap.s1/s2/s3.grle`. It is anchored at both ends, so separators, `..`, prefixes and suffixes are refused.
  - `fileExists(path) == true`
  - a literal `true` engine load
  - the width match, when a width is expected
- **The tools follow the width.** A successful load that changes the width rebuilds the modifier and filter against the loaded map.
- **It reports whether it attempted an engine load**, so the caller decides what an attempted failure means.
- `adoptNativeGeneration` passes `envelope.filename`. It no longer reopens by generation number.
- **The legacy import moves into the barrier's legacy branch**, through a new `ctx.legacyProbe` that `tryRestoreMissionWater` supplies. The branch calls it only when `type(ctx.legacyProbe) == "function"`, and only on a **pre-generation** load.
- **The generation-era test** is `SaveLoadHandler:isGenerationEraLoad`. A load counts as generation-era when:
  - any COMPLETE candidate from either mirror was collected, valid or not, **or**
  - either staged snapshot carries `saveGeneration` above 0.
  - It never uses the selector's mode, nor the legacy branch's single snapshot.
- **A generation-era load with no usable envelope declines the live map to ZONE** (`result.declined`). It does not import the legacy image and it does not seed a fresh fine map.
- **Legacy outcomes differ.**
  - An absence, or a refusal before any engine load, does nothing.
  - An **attempted** load that fails resets the map fresh at the computed resolution and rebuilds the tools **unconditionally**, even when the width did not change.
  - If even the reset fails, the map is released and the soil system declines it.
- **A nil save directory** makes every probe and open a no-op.
- **With the release gate off** there is no map, so nothing is imported or declined. Bob's intake note: `legacyProbe` and `loadNativeFile` are no-ops there.

### 2. Three rotating slot files
- The image lives in `csMoistureMap.s1.grle`, `s2` or `s3`. The generation stays in the envelope.
- **The cut chooses the slot once**, before the native write. It takes the lowest slot named by neither the selected current nor the selected previous retained record.
- **That one string is used everywhere:** `capture.filename`, the `saveNativeMap` argument, the `saveToSavegame` argument, the written path and the path the cut checks on disk.
- `generationFileName` is removed. `saveToSavegame` refuses any name that is not a slot.
- The cut keeps its `sgDir ~= nil` guard. It also requires a chosen name before calling the writer.
- **No file is deleted.** Old `g<N>` files are never written again. They still load when an envelope names them.

### 3. Exact-true registration
`CropStressStateLedgerBridge.register` sets `active` only when the `pcall` succeeded **and** `registerModule` returned exactly `true`, the rule the NetworkSync bridge already uses. The log says "registration failed or refused" with the error or the return value. `delivered` and `pendingState` handling is unchanged, including a synchronous late delivery.

## Named limits carried from the certified design (not fixed here)
- **A legal name can stand in for another.** The compact digest does not cover `filename`, so the door check cannot catch the swap.
- **Lagging mirrors are not protected.** Only the files named by the winning mirror's current and previous records are. A name held only by a lagging mirror can be rewritten.
- **A save that lost both whole moisture records** reads as pre-generation and can import a stale legacy image. Nothing on disk separates it from a pre-generation save interrupted after its first slot write.
- **Saving after a ZONE decline:** the next load seeds a new fine map from saved averages. The older fine pair is not reselected.
- **Leftover PR188-era `g<N>` files** stay on disk. The mod cannot remove them. The tester release note is Sasha's.
- **The career XML handle** (observation 7) is observation only.

## Tests
- **New spec** `tools/test/lua/RSF-F244-moisture_generation_cleanup_spec_test.lua`, **165 assertions**.
  - **It drives the real code:** the value map, soil system, save handler, manager barrier and StateLedger bridge.
  - **The engine is stubbed at its API:** `createBitVectorMap`, `loadBitVectorMapNew`, `loadBitVectorMapFromFile`, `getBitVectorMapSize`, `saveBitVectorMapToFile`, `fileExists`, `DensityMapModifier.new` and `DensityMapFilter.new`, each recording what it was asked.
  - **None of this ran under test before.** The existing benches replace the value map with a fake, so no bench had executed the loader, the probe or the slot write.
  - **The groups:**
    - P: the probe opens nothing, including with the gate off.
    - N: the loader's door, disk check, receipt, width and tool rebuild.
    - L: legacy absence, attempted failure and reset, success, and a failed reset.
    - G: the soil system seams.
    - B: the barrier's legacy branch. It covers a pre-generation import through the manager, generation-era declines by invalid envelopes, by `saveGeneration` on either mirror and by invalid envelopes alone, the gate-off case and a missing callback.
    - W: one name end to end, the s1 s2 s3 s1 s2 rotation, an interrupted write and its retry, a nil directory, and the bank after a restore.
    - R: exact-true registration, including a synchronous late delivery.
  - Every case runs inside `group()`, so a Lua error fails a named row instead of the suite discarding the file.
- **Updated benches.** `scs039_native_generation_test` (63) and `scs039_save_view_restore_test` (188) move from generation names to the slot scheme.
- **Suite:** 2371 passed, 0 failed across 55 files. The syntax gate is clean.
- **Mutations: 22 run, 22 killed.** Each was a literal byte replacement asserted to match exactly once, re-read to prove it landed, then restored and sha256-compared.
  - **Name grammar:** `g0` and leading zeros accepted (MA); end anchor dropped (MB).
  - **Refusals removed:** the loader's door check (MC); the writer's slot-only check (MD).
  - **Slot choice** ignores the previous record, in the value map (ME) and in the cut (MO).
  - **Tools** not rebuilt on a width change (MF).
  - **Legacy reset** removed (MG), or applied only when the width moved (MH).
  - **The probe** marks the fresh map loaded (MI).
  - **Generation-era test:** invalid COMPLETE rows ignored (MJ); only the first surface walked (MK); no decline (ML); a decline with the gate off (MM); the legacy probe also runs on a generation-era load (MN).
  - **The nil-directory guard** removed from the cut (MP).
  - **Adoption** reopens by a generation-derived name (MQ).
  - **The legacy import** stops being a no-op without a live map (MR); a map that could not be reset is not declined (MS).
  - **The manager** supplies no `legacyProbe` (MT).
  - **Registration** trusts any non-throwing call (MU), or never captures the result (MV).
  - **Stated plainly:** MJ SURVIVED the first run. Every invalid-envelope fixture also carried the flat `saveGeneration` key, which marks the load by itself. Group B32 removes that key, and MJ now dies on B33 and B34. MD first died on a load error rather than a name; the write section of the native-generation bench now runs under `pcall`.

## For F245/F247 (Bob's intake note)
The install branches F247 hangs its carrier-currentness record on are unchanged in shape, at their new coordinates in `SaveLoadHandler.lua`:
- the envelope branch at `:1484-1522`
- the legacy branch at `:1547-1578`

The legacy branch gains only the generation-era test and the `legacyProbe` call at its top.

## Not checked
- **In game.** Deployed with the game not running; `log.txt` dates from 2026-09-15. None of the 13 native observations in the brief has been run. The TESTING.md row carries them.
- **`saveBitVectorMapToFile` over an existing file:** whether it overwrites atomically was not examined.
- **Engine validity of a modifier** after a size-changing load is not determinable from Lua. The code rebuilds instead of relying on it.
- **Multiplayer and dedicated server.** Not run. The cut and the barrier are unchanged in where they run.
- **`SaveLoadHandler.lua` is 1635 lines**, over the repo's 1500-line rule. It was already 1581 before this PR. It was not split here, because F245/F247 build on these coordinates and a split belongs in its own PR.

## Build note
`bash build.sh --deploy` names its zip after the folder when `zip` is not installed, so from a worktree it built `_wt-f244.zip`. Worse, the deploy step deletes the installed `FS25_SeasonalCropStress.zip` before its copy, so that copy failed with no mod left installed. I deployed the same built zip under the right name, minus the worktree's one-line `.git` pointer file, and removed the misnamed zip. The script is unchanged here; that belongs in its own chore.

## Commits
- `65a7799` feat(RSF-F244): moisture generation cleanup: no-open probe, three slot files, exact-true ledger registration

Manual test row: TESTING.md in ecosystem-dev-tracking (RSF-F244).


TheCodingDad-TisonK:
## Bob cold review, head 65a7799 (matches ls-remote), base development 4c144b4

**Verdict: PASS.** No BLOCKER and no MAJOR. Three MINORs, none of which blocks the merge.

Commits on the branch (1):
- `65a7799` feat(RSF-F244): moisture generation cleanup: no-open probe, three slot files, exact-true ledger registration

The title matches it.

I checked the diff against the certified F244 brief (tracking e38cda7), fix items 1-3 and its invariants.

### Findings
- **src/SaveLoadHandler.lua: MINOR. The file is now 1635 lines, over the 1500-line rule.** Deferring the split keeps F247's coordinates stable, and I accept that. Open a split as a follow-up after F245/F247 merges, not inside that pair.
- **build.sh: MINOR, outside this diff.** Run without an explicit zip name, it names the zip after the folder (`_wt-f244.zip`) and deletes the installed SCS zip before a `cp` that then fails. A worktree build therefore uninstalls the mod. This needs its own chore PR.
- **src/SoilMoistureSystem.lua:329 (`initValueMap(savegameDir)`): MINOR.** The parameter is now unused. Drop it or comment it when that line is next touched.

### Rulings on the two choices
1. **`result.declined` only when `soil:mapActive()`: accepted.**
   - With the gate off there is no map for the decline step to release, and the brief's invariant says the cell-store path is unchanged.
   - B25/B26 pin this: no decline, and nothing opened.
2. **Gate-off `legacyProbe` no-op through `importLegacyNativeMap`'s `mapActive` check: accepted.**
   - With `valueMap` nil the loader is unreachable.
   - B27/B28 pin this.

### Holds, checked against source
- **Probe (item 1).**
  - `initialize()` opens no file and leaves `loadedFromSave` false.
  - The only opener is `loadNativeFile`, reached from `adoptNativeGeneration` (envelope.filename) and `importLegacyFile`. A grep at the head finds no remaining `loadGenerationFile` or `generationFileName` reference.
- **Door.** The check is an anchored whole-string match: legacy, `^csMoistureMap%.g[1-9]%d*%.grle$` (refuses g0 and leading zeros), or s1-s3. It is followed by `fileExists == true`, a pcall with a literal-true result, and the width check when a width is expected.
- **Attempted versus absent.**
  - A refusal, a missing file or a nil directory returns `false, false`, with no reset (L1-L5).
  - An attempted failure resets to `computedResolution` and rebuilds the tools unconditionally (L6-L13), including a partial load that moved the width (L11).
  - If even the reset throws, the map is deleted and `importLegacyNativeMap` declines it (L19-L21). That decline runs before the seed pass, so nothing seeds a released map.
- **Tool width.** A successful load at a new width rebuilds the modifier and filter and records `toolWidth` (L14-L17). A failed rebuild returns `false, true`, so adopt declines and legacy resets.
- **Generation-era test.** `isGenerationEraLoad` walks every row `collectMoistureCandidates` returns; that collector appends COMPLETE rows whether or not `compactOk` passes. It also walks both staged surfaces for a numeric `saveGeneration > 0`.
  - `STAGE_ORDER` is defined at :67.
  - The XML reader takes `#saveGeneration` via getInt.
  - StateLedger round-trips numbers typed (StateLedgerXML.lua:82, :136).
  - B32-B34 isolate the invalid-COMPLETE-rows-only case with the flat key removed.
  - The test is not selector mode and not the single `snap`.
- **Write path (item 2).**
  - `chooseSlotFileName` is a pure function of the envelopes' filenames in `_completePair.current/previous`.
  - That one string is passed to `saveNativeMap` and `saveToSavegame`, and checked with `fileExists`.
  - `saveToSavegame` refuses non-slot names.
  - The `sgDir ~= nil` guard is kept and `filename ~= nil` is added (W18-W20).
  - Rotation, bank protection and file count are covered by W1-W10.
- **StateLedger (item 3).** `registered` is captured inside the pcall; `active` is true only when `ok and registered == true`, and is set false otherwise. The synchronous late-delivery handling is unchanged.
- **W11.** It is a fair model of what the cut can detect: the engine reports true, the file is not on disk, nothing commits, only the candidate slot was touched (W12-W14), and the retry reuses that slot (W16). It does not model a partial file that exists on disk after a crash. That is PR188's receipt limit, not this repair's, and the named bank guarantee still holds because the candidate slot is unnamed until commit.

### Not checked
- I did not rerun the suite (2371/0) or the 22 mutations.
- The native engine behaviour is unverified (brief observations): a reload of `loadBitVectorMapFromFile` into an existing bvm, tool validity after a width change, and slot writes on disk.
- I did not trace the other callers of `restoreMissionWater` beyond the manager.
- The TESTING row exists (tracking 253ab60). I read only its opening.

Bob, engineering seat

sasha-rf (approved):
﻿RSF-F244: moisture generation cleanup (no-open probe, three rotating slot files, exact-true StateLedger registration). Head 65a7799e matches origin.

Fixes three defects: a stale legacy moisture image could survive as current truth on the restore barrier's no-candidate branch; every save left the previous moisture image on disk permanently (unbounded growth, the mod cannot delete these files per the engine's deleteFile sandbox); and the StateLedger bridge reported a refused registration as registered by ignoring registerModule's boolean return. Fix moves to a single loader with an anchored slot-name grammar (csMoistureMap.grle / .g<N>.grle / .s1-s3.grle), a fixed three-slot rotation chosen once before the native write, and exact-true registration matching the NetworkSync bridge's existing rule.

Everything native here sits behind the cs_grid_concordance release gate except the StateLedger fix, which applies to every save regardless of settings. 165 new spec assertions, 22/22 mutations killed (two fixture gaps, MJ and M8/M16, found on first run and closed), full suite 2371/0, syntax gate clean. Bob PASS (6051b64) with MINORs that don't block. Named limits carried from the certified design (legal-name swap, lagging mirrors, zone-decline reseeding) are documented, not fixed here, and match the certified handoff's scope.

TESTING.md row exists in ecosystem-dev-tracking (line 43, RSF-F244) but carries no recorded observation yet: thirteen in-game passes written, explicitly unobserved. That does not block this PR's merge into development; it blocks release triage, and is Tyson's to close.

Approving as sasha-rf: code and test coverage are clean.

UTF-16 scan: 8 file(s) read at the PR head, all clean.

PR #192, opened 2026-09-17
RSF-F245 + RSF-F247: moisture restore honesty and field ground recovery

TheCodingDad-TisonK:
## RSF-F245 + RSF-F247: moisture restore honesty and field ground recovery

One implementation unit, as the delivery defines it: F245 (restore and read honesty, brief v0.7) and F247 (membership and geometry recovery, brief v0.5), built against the certified paired contract model. Base: `development` at 4a2b2f3 (#190, F244).

### Review round 1 (Bob's cold review of 61ea779, ledger 2d5ca96): now at c2d16c8

**BLOCKER, pure-client desync: fixed with a server gate.**
- `refreshForPublication` returns at once on a pure client (`g_server == nil`).
- `_refreshFieldAggregate` never marks a field unavailable on a client:
  - OK with a mean still makes the field current and clean;
  - a refusal still fails the client's provider closed;
  - every other outcome leaves the host's number in place, exactly as before F245.

**Why the gate and not the receive paths.** F245 is host-scoped: the brief puts client truth in RSF-F249 (:155, :186, observation 7 at :288), and a client "keeps whatever number it last had". Marking fields current in `CropStressMoistureInitEvent:run` and the NetworkSync receive would have started building client availability state inside this pair. It would also still blank a field between its enumeration and the next sync.

**Named rows, Bob's probe shape** (g_server nil, a live map with no pixels, a dirty field, the host value delivered around the HUD rebuild):
- C1: the init event.
- C2: the NetworkSync mirror, with the map live but not current, so the mirror applies.
- C3: reached control. The host still marks the same shape unavailable.
- C4: a client refusal fails closed and keeps the host's number listed.

All 18 new named rows in this round **fail on 61ea779** and pass at c2d16c8. I checked this by running the new specs against a detached checkout of 61ea779.

**MAJOR 1, call site b: conformed to the model.** Refresh and the daily settle reach the ground check only on `EMPTY` (model :97). OK with no mean, or an unknown outcome, is unavailable with no probe.
- H8 (refresh) and D4 (settle), ready with a carrier recorded, assert no probe, no fill and no decision.
- Each has a reached control on the same fixture that does probe on EMPTY.
- Bob's M-B, reversed (the wider trigger restored): MB1 is killed by H8c-f and MB2 by D4b-c.

**MAJOR 2, native read cost (named, behaviour unchanged).** A field that is unavailable stays dirty, so every publication door re-reads its first polygon natively; a field with a current value costs nothing.
- **Correction to the one-read-per-second estimate.** `HUDOverlay:update` throttles its list rebuild with `rebuildTimer >= 1.0` (HUDOverlay.lua:300-303), but `dt` is in milliseconds (BaseMission.lua:770 adds it to mission time). So while the moisture HUD is visible on the host the rebuild runs **every frame**, and has done so since before this pair (the same line is at 4a2b2f3).
- **HUD cost on the host:** one native polygon read per blank field per frame, about 60 reads per second per blank field at 60 fps. If the HUD's selected field is blank, the forecast strip's `getMoisture` adds another read per frame.
- **Other doors** each cost one read per blank field per call: the PDA table and stats, guest PDA, the consultant list and recommendation, the hourly act and the stress pass.
- **Door B re-walk:** a field whose outline was refused or partial is re-collected at most once every 5000 ms, when a door asks for its polygons. The collection walks the engine field list and each matching field's nodes and engine polygon. An INVALID_FIELD_GEOMETRY field does no native read, only that re-walk.
- **Held with Tyson, not changed here:** whether a read may be skipped until `moistureRevision` moves is a Design question. So is fixing the HUD throttle unit, which is pre-existing and would change the read cadence.

**MAJOR 3, doc.** A native refusal during the daily settle marks only the field being read as unavailable, fails the provider closed and stops the walk. Fields later in the walk keep their numbers from before the refusal, as the provider's existing degrade does.

**MINOR.** A new ZONE cell on a field with no current value no longer starts from an invented 0. The water stays field-wide pending, like the cap branch (row W4, mutation Z01).

**Checked, per Bob:** the dd22531 refusal reading follows model :94-97. The cooldown order is correct. The render paths read nil-safe; Bob did not run them.

**Round 1 numbers:**
- Suite 2861/0 across 60 files.
- Mutations 37 of 37 killed: the first 31, all re-run against c2d16c8, plus MB1, MB2, the client gates CG1-CG3 and Z01.
- CG3 (a client refusal marking the field unavailable) survived at first. Row C4 was added and it is now killed.
- Installed zip sha256: before `ab578a35e293bcfad45df33630b1472041e3129bc0eb8710e4baf7ced1c04085` (the #193 build), after `7c46c31c5a827708d8f2b627993ff667eaa1442699ef746ab537401a6198c4a2` (c2d16c8, equal to the built zip, 172 entries, changed sources byte-identical).

**Not checked this round:**
- an in-game run, including the per-frame read cost on a real map;
- a real client join;
- whether any other client path marks a field unavailable. `migrateFieldToMap`'s pre-ready "no base" branch and the marking pass are not reached on a client by any caller I read, but I did not trace every client-side caller.

### What the pair does

**F245: a field with no current moisture has no number.**
- Every field record has `aggregateState` (CURRENT or UNAVAILABLE) and a reason. While unavailable the current slot is empty; the last value is memory only (`moistureLastKnown`), never saved, sent or read as current.
- `_refreshFieldAggregate` and the daily settle have typed outcomes and no silent return: OK with a mean is current and clean; EMPTY, OK with no mean and missing geometry are unavailable and stay dirty; a refusal fails the provider closed.
- `getFieldAggregate` never answers a retained cell mean while the map is active. The seed base never falls back to 0.5.
- A map loaded from its own save gets a pure marking pass (no paint, no native read). A fresh map seeds from the fallback base; a field with no base is not painted.
- `refreshForPublication()` runs once before every host door: the hourly loop, the sorted list, PDA table and stats, guest PDA, consultant list, band check and recommendation, stress modifier, both client snapshots.
- Restore: a row with no value installs nothing (no `clamp01(nil)`); installs mark the field current and record SAVED or FRESH_START; ZONE fields settle to their cell mean after the decline step. The XML reader tests attribute presence (`XMLFile:hasProperty`, decompiled xml/XMLFile.lua:69). Capture, own XML, the ledger row, NetworkSync serialize and the init event all leave an unavailable field out; both receives never write a non-number.
- Consumer floor: nothing compares, ranks, averages, formats or alerts on a missing moisture. Rows stay, cells go blank, sorts put no-reading rows last with no `or 0`.

**F247: water and reads land inside the field, and blank ground is recovered once.**
- `CropStressValueMap:hasWrittenPixels` (filtered BETWEEN RAW_MIN..RAW_MAX) and `fillUnwrittenPolygon` (filtered EQUAL 0, decided by a falling count, never by set success).
- Two-source parcel collection: the node walk is preferred, the engine polygon (`getDensityMapPolygon`, else the `densityMapPolygon` field) is used when the walk fails; a parcel with a failed field is a partial entry.
- Retry doors: A (ownership change, unchanged), B (5000 ms expiry, every refused or partial collection restamps `refusedAt`), C (hourly, once per hour key, server only). A changed fingerprint reopens only `_groundChecked`, never `_mapSeeded`.
- `_checkFieldGround` implements the whole decision table with the carrier gate (SELECTED_PAIR, LEGACY_IMPORT, FRESH). Post-ready migration is only the check; it never reaches the unfiltered paint.
- Membership at `getMoisture` (positional) and `applyWaterAtCell`, before the fail-closed test and before any routing. `_memberAnchor` for the two display-centre irrigation calls; no anchor means the call is skipped with one log line.
- The capture reads the envelope revision after its refresh loop. Teardown empties every new table.

### New file
`src/SoilMoistureGround.lua` (279 lines): parcel probe and fill, the ground check and seed, and the member anchor. `main.lua` sources it right after `SoilMoistureSystem.lua`.

### Design answers folded in (Iris, ledger a2ae501)
1. **Door C:** explicit `g_server` guards inside `_retryRefusedGeometry` and `_checkFieldGround`; the `hourlyUpdate` call site keeps its own guard too.
2. **`aggregateDirty`:** the initial meaning is dirty. `enumerateFields` sets `aggregateDirty = true` on record creation, and `_refreshFieldAggregate` returns on `~= true` (nil is clean, as in the certified model at :33, :92, :224). Named rows cover **both doors**: E1d and E3a (publication), E4a and E4c (capture).
3. **Missing `readAverageOfPolygon`:** a provider capability refusal. Refresh and settle test for the typed read before any geometry and fail closed. The field is also marked unavailable with reason `PROVIDER_REFUSAL` on every refusal, so the stale scalar is never kept as current. The brief lists three reasons; the certified model's refresh uses the outcome itself (`markUnavailable(f, r.outcome)`, model :94-96), so this follows the model. Named rows H6b, H7a, H7b, D3b.
4. **Overlay:** see player-visible consequences.
5. **F242/F249:** not delivered; see below.

### Player-visible consequences (for the release note)
- **Overlay:** off-parcel samples now refuse, and the overlay can fall back to the field aggregate. That is the accepted F247 result. "Unchanged" meant no F242 redesign in this pair, not identical pixels.
- **Blank values:** a field with no current moisture shows blank moisture cells, no percent, no severity colour and no bar in the HUD, PDA, guest PDA, consultant and schedule dialogs. It is left out of averages, driest lists and recommendations. No new strings.
- **F242 and F249 are NOT delivered and do not ship before this pair.** Sasha's release note carries the path-dependent overlay picture (pixels, hole, or the good green) and the client stale or start value for rows the server now omits. A client keeps whatever it last had, or its enumeration start value, for such a row.

### Bob intake (ledger 35d585e): watch list
- **Readers missing from the floor:** HUDOverlay :1133-1135 (bar and colour) and :1587-1589 (stub rows) are fixed. HUDOverlay :1399-1403 was already nil-safe at base; its sort comparator is fixed. CsPDAScreen :488 (merge minimum) and :618-620 (average) are fixed. CsRfPdaGuest :324-376 (next-step ladder) is fixed. CsRfPdaGuest :1100-1104 and :2222-2226 read `topRiskFields`, which now leaves out a field with no reading. The client new-entry writes in the init event and NetworkSync receive never write a non-number.
- **ZONE PENDING_ONLY install** marks the field current.
- **The fourth silent return** is now the capability refusal (Design item 3).
- **Delta filter wording:** row S1 has a delta change the compare values before a probe, and the probe still counts BETWEEN 1..255.
- **Early-exit trap:** every "no number" row asserts its `readAverageOfPolygon` count. Every ground-check row asserts its parcel probe count (G*p rows: 0 when refused, partial or not run, 1 when decided). The reload case does not pre-set `_mapSeeded` (X5).
- **Filter stub:** real EQUAL, BETWEEN and NOTEQUAL over a pixel array, a set that writes nothing, and throw and nil hooks.
- **fengari pairs:** the publication walk and door C walk sorted ids. The consultant list ends in a total-order sort (risk, then field id). Rows P1b and R5d say that the bench is insertion ordered and cannot show hash-order nondeterminism.
- **Runoff fence:** NOT checked. `runoffSystem:acceptSurplusSpan` was not read. Membership is enforced in `applyWaterAtCell` before numeric and span routing.

### Other named choices
- **Existing fixtures changed:**
  - `scs039_delegation_test`: the settle EMPTY row now expects unavailable, nil moisture and last-known 0.42.
  - `scs039_iris_review_fixes_test` and `scs041_wiring_test`: a `{0..100}` outline so their positional writes pass membership.
- An envelope-only field (no own-XML row) counts as a saved row for the ground check.
- The carrier record is computed after the decline step.
- `CropConsultant:onCriticalThreshold` now tests for a number before stamping the cooldown, so an event with no number never uses up a field's cooldown.
- **1500-line rule, flagged, not split:** `SoilMoistureSystem.lua` 3639 lines, `SaveLoadHandler.lua` 1735, `HUDOverlay.lua` 1760, `CsRfPdaGuest.lua` 2832. The new ground-check code went into its own file; splitting the existing files is out of scope for this pair.

### Tests
- Suite at 61ea779: 2833 passed, 0 failed across 60 files (2371/0 across 55 before the pair). At c2d16c8: **2861/0/60**, see Review round 1.
- `tools/test/lua/f245_harness.lua` runs the REAL `CropStressValueMap` over a pixel-grid modifier and filter. There are five specs, each group under pcall:
  - `RSF-F247-ground_check_spec_test.lua`
  - `RSF-F247-retry_membership_anchor_spec_test.lua`
  - `RSF-F245-aggregate_honesty_spec_test.lua`
  - `RSF-F245-restore_capture_transport_spec_test.lua` (through the real restore barrier)
  - `RSF-F245-consumer_floor_spec_test.lua`
- The Lua 5.1 syntax gate was proven on a broken control file first.
- **Mutation pass: 31 of 31 killed by named rows.** Every anchor matched exactly once, the landed edit was proven, the full suite ran, and each restore was sha256-checked.
  - **Bob's 17:** B01 cell mean while active, B02 EMPTY keeps scalar, B03 marking pass paints, B04 post-ready fall-through to paint, B05a/b/c unfiltered set, probe count and fill count, B06 OK on set success, B07 PARTIAL not closed, B08 `_groundChecked` never set, B09 reopen clears `_mapSeeded`, B10 `refusedAt` not restarted, B11 membership after the pending store, B12 empty carrier seeds, B13 capture revision before refresh, B14 teardown misses a table, B15 `or 0` in the comparator, B16 serialize sends 0.0, B17 door C on a client (inner guard, and both guards).
  - **Extras:**
    - E01 enumeration not dirty (kills Bob's E1d plus Design's E3a, E4a, E4c) and E02 nil treated as dirty
    - I01 missing read silent, I02 and I03 refusal keeps scalar
    - P01 unsorted publication walk
    - R01 row installs 0, R02 no XML presence test, R03 carrier ignores `loadedFromSave`
    - A01 anchor takes an outside centre
    - C01 cooldown before the number test

### Not checked
- **No in-game session.** log.txt is from 2026-09-15, so there is no `[CropStress]` evidence for this build. The TESTING.md row carries the in-game pass.
- **No native observations.** F245 1-14 and F247 1-17 were not run. What the native unfiltered count means is not determinable (F250 item 3); the harness models written pixels only, and F247's decisions never use it.
- **Hard limit, as the brief states:** on a tolerated-error parcel with several outlines, the host number is still the first-polygon read.
- **Consumer floor not reached on the bench** (render or local functions with no GUI double):
  - HUDOverlay drawing and grouped sort
  - IrrigationScheduleDialog labels
  - the settings driest list
  - the consultant dialog list rows
  - the guest PDA field rows, `topRiskFields` and the detail band

These were reviewed in the diff only.

### Deploy
`bash build.sh --deploy` was not used. This machine has no `zip` binary, so from a worktree the script's Python fallback writes `../_wt-f245.zip` while the deploy step deletes the installed zip and then copies `../FS25_SeasonalCropStress.zip`, which does not exist. Instead I ran `bash build.sh`, checked the listing (172 entries; modDesc, main and every changed source present and byte-identical to the worktree; no `tools/`; forward slashes) and copied the zip in.
- Installed zip sha256 **before** `4ed6580b04d99628eca8edcc8494c661e1e31f27b81d07829752e686d0e81b13`
- **after** `e69104f3461db198ee34d0013e3f29da4ab0a64a94dfb6b23620886bbc4d5f1a` (equal to the built zip)

The worktree's `.git` pointer file is packed into the zip, the same defect class as the MarketDynamics and FarmTablet chores; it goes into `chore/build-sh-safe-deploy`.

### Commits
- b3d1eea aggregate honesty core, parcel probe and fill, field ground check, geometry retry doors
- 51de721 restore, capture and carrier record keep absence absent
- bf2535a snapshot refresh, debug driest list, member-anchor irrigation calls
- d539446 host consumer floor
- dd22531 a provider with no typed read is a capability refusal (Design a2ae501)
- 7b28b0f pixel-grid bench for the pair
- 61ea779 parcel probe count on every ground-check row; no-carrier migration never paints
- c2d16c8 review round 1: client keeps the host's number; call site b only on EMPTY; no ZONE 0 cell

The FarmTablet alert-loop guard shipped separately as #163.


sasha-rf (approved):

UTF-16 scan: 27 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #192 (RSF-F245 + RSF-F247)

**Head checked:** `git ls-remote` branch head = `refs/pull/192/head` = `gh` headRefOid = `61ea779`. Base `development` `4a2b2f3`. Full diff read (27 files, +3349 / -348).

**Verdict: FAIL** (one BLOCKER, three MAJORs)

### Findings

- `src/SoilMoistureSystem.lua:1345` (with `src/events/CropStressMoistureInitEvent.lua:107`, `src/integrations/CropStressNetworkSyncBridge.lua:226`): **BLOCKER**: pure-client MP desync. `refreshForPublication` has no server gate, and `initValueMap` stands a live map up on clients too. On a pure client, `enumerateFields` creates every record dirty, and the 1 Hz HUD rebuild (`HUDOverlay.lua:303` -> `getFieldsSortedByMoisture`) refreshes it against the client's local map. Where that map has no pixels for the field, the field is marked `UNAVAILABLE:EMPTY`. Both receive paths then write the host's number into `existing.moisture` but never touch `aggregateState`, so the HUD, PDA and consultant show no reading while the host publishes one. The map only reaches clients at join (`CropStressManager.lua:1092` `queueMapSync`, no later deltas), so any host seed after a client joins stays blank on that client for the whole session: a geometry retry through door B or C, a fingerprint reopen, or a call site b seed on the first hour. Before this PR the client list read `data.moisture` directly. Proven on a scratch copy of 61ea779 with a probe row (g_server nil, live map with no pixels, dirty field, `CropStressMoistureInitEvent:run` delivers 0.55): the slot takes 0.55, and `getFieldsSortedByMoisture` still returns nil (C3 FAIL). **Fix:** return early from `refreshForPublication` and from the unavailable marking in `_refreshFieldAggregate` when `g_server == nil`, or make both receive paths mark the field current and clean. Add a named pure-client row covering both receive paths.
- `src/SoilMoistureSystem.lua:1402` and `:2726`: **MAJOR**: call site b fires on every non-OK outcome that is not a refusal or bad geometry, including OK with a nil mean and any unknown outcome. The certified model fires it only on `outcome === "EMPTY"` (`f245-f247-paired-contract-model.mjs:97`). An OK-with-nil read (pixels that decode to no data) reaches the probe, and the probe filters `RAW_MIN..RAW_MAX`, so it can answer NONE and fill that field. **Mutation M-B** (gate both sites on `outcome == "EMPTY"`) **SURVIVED**: the bench was green with the hunks confirmed applied, so the bench cannot tell this code from the model. **Fix:** gate both sites on `outcome == "EMPTY"` and add a row with `ready = true` asserting that OK-with-nil runs no probe and no fill. The alternative is to get Design to bless the wider trigger.
- `src/SoilMoistureSystem.lua:1345` (the publication doors): **MAJOR**: an unavailable field stays dirty for good (as the model requires), so every door re-runs `readAverageOfPolygon` on it: the HUD rebuild at 1 Hz, NetworkSync `_onWriteState` at 1 Hz, and the PDA, guest PDA, consultant, dialog and hourly doors. An `INVALID_FIELD_GEOMETRY` field also re-walks its engine nodes every 5 s through door B. That makes two or more native polygon reads per second per blank field, with no bound, on a host with the HUD open. It is not a behaviour defect. **Fix:** name the cost in the PR body, or ask Design whether a read may be skipped until `moistureRevision` moves.
- `src/SoilMoistureSystem.lua:2721`: **MAJOR** (doc): after a settle refusal, `break` leaves only the lowest-id dirty field blank. Every later field keeps its number and publishes under ZONE rules for the rest of the mission. That matches the model's one-way path, but players will see it (one field blank, the rest numeric) and neither the PR body nor the TESTING row says so. **Fix:** one sentence in the PR body and TESTING row 49.
- `src/SoilMoistureSystem.lua:1805`: **MINOR**: the ZONE cell creation in `_rawWaterStore` still seeds `or 0` when the field has no value. The map branch 60 lines above was changed so it no longer invents 0. It is only reachable on a field left unavailable after a fail-closed, where the cell's 0 later enters the ZONE cell mean once `_writeFieldMoisture` makes the field current. **Fix:** keep the gain pending the same way, or skip the cell.

### The checks Fred asked for

1. **dd22531 reading: CONFIRMED.** The model's refresh calls `markUnavailable(f, r.outcome)` for every non-OK outcome, refusals included (`paired-contract-model.mjs:94-97`), and Iris's "never keep the old scalar as current" agrees. The capability refusal is tested before any geometry (H7c: walks 0).
2. **`onCriticalThreshold` number test before the cooldown stamp:** correct. An event with no number does not use up the cooldown.
3. **Render paths the bench cannot reach:** read in HUD, PDA, guest PDA, consultant dialog, schedule dialog and settings panel. Every comparator is nil-safe with ties broken by field id, and none of the render paths I read still has an `or 0` / `or 0.5`. I did not run any of them.
4. **Deploy:** out of scope for this PR. The `build.sh --deploy` hazard (it deletes the installed zip, then the copy fails when no zip binary is present) belongs to the SCS build.sh chore.

### Earlier intake commitments

- `enumerateFields` sets `aggregateDirty = true` (`:932`), and `_refreshFieldAggregate` returns on `~= true` (`:1372`). **Mutation M-A** (set it to nil) is **KILLED** by E1d, E3a, E4a (plus E1b/e/f, E3b).
- Door C (`:1010`) and `_checkFieldGround` (`SoilMoistureGround.lua:120`) carry their own `g_server` guards. **Mutation M-C** (remove the check's guard) is **KILLED** by G21a/G21b/G21p.
- Design a2ae501 item 2: E3a (publication door) and E4a/E4c (capture door) cover both doors for unpainted and painted new fields.

### APIs verified against the decompile

| symbol | where | note |
|---|---|---|
| `modifier:executeGet(filter)` returns `(_, numPixels, _)` | `field/FieldManager.lua:100` | filtered count |
| `modifier:executeSet(value, filter)` | `densityMaps/DensityMapHeightUtil.lua:332`, `environment/SnowSystem.lua:246` | |
| `filter:setValueCompareParams(DensityValueCompareType.EQUAL / BETWEEN, ...)` | `PlaceableHusbandryMeadow.lua:344`, `:433` | |
| `XMLFile:hasProperty` | `xml/XMLFile.lua:69` | |

### Bench

`node run-tests.mjs` at 61ea779 in the worktree: **2833 passed, 0 failed across 60 files**. The worktree was left clean. The mutations and the client probe ran on a scratch copy outside the repo. In that copy `scs018_decay_catchup_test.lua` cannot load, because the copy has no sibling `FS25_TimeGuard`; that is an artifact of the copy, not a failure.

**Not checked:** any in-game or dedicated-server run, native observations, `runoffSystem:acceptSurplusSpan`, Fred's 31-mutation harness (I did not re-run it), the guest PDA `topRiskFields` path at `:1100` / `:2222` beyond reading it.

### Commits on the branch (7)

```
61ea779 test(RSF-F247): parcel probe count on every ground-check row; no-carrier migration never paints
7b28b0f test(RSF-F245/F247): pixel-grid bench for the moisture honesty pair
dd22531 fix(RSF-F245): a provider with no typed read is a capability refusal
d539446 feat(RSF-F245): host consumer floor never compares, ranks, averages, formats or sends a missing moisture
bf2535a feat(RSF-F245/F247): snapshot refresh, debug driest list, member-anchor irrigation calls
51de721 feat(RSF-F245/F247): restore, capture and carrier record keep absence absent
b3d1eea feat(RSF-F245/F247): aggregate honesty core, parcel probe and fill, field ground check, geometry retry doors
```

The title "RSF-F245 + RSF-F247: moisture restore honesty and field ground recovery" matches all seven commits. TESTING.md row 49 exists and names concrete in-game checks. Once the BLOCKER is fixed, its dedicated-server line ("client HUD and PDA show the same values as the host") becomes a real check.

Bob

sasha-rf (approved):

Fresh review at c2d16c8 (review round 1). Read the full CropStressMoistureInitEvent:run and CropStressNetworkSyncBridge._onReadState client receive paths, not just the diff, per Desk's ask (same class of gap as #965). Confirmed: (1) refreshForPublication now returns 0 immediately when g_server==nil, and _refreshFieldAggregate's fallback elseif also refuses to mark a client-local field UNAVAILABLE - both receive paths (init event and NetworkSync mirror) only ever write .moisture/.stress, never touch aggregateState, so a client keeps whatever CURRENT/UNAVAILABLE state it had and the host's number is never blanked by a client's own empty map pixels. (2) The EMPTY-only ground-check gate (call site b) is applied at both call sites: _refreshFieldAggregate and the inlined settleDaily branch. (3) settleDaily's un-guarded EMPTY-marking is safe because every call path into it is itself g_server-gated (onHourlyTick's g_server block, and Time Guard's registerAccrual only registers when g_server ~= nil) - so it never runs on a client despite lacking its own guard. (4) The water-seed MINOR (field-wide pending instead of an invented 0) is in place. Ledger 2c2a0ff and TESTING row 49 confirmed on origin/main. Approved, awaiting Bob's re-review of this head.

UTF-16 scan: 27 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
## Bob: re-review, PR #192 round 1 (c2d16c8)

**Head checked:** `git ls-remote` `refs/pull/192/head` = `gh` headRefOid = `c2d16c8`, one commit on top of 61ea779. I read the delta, then the whole of every function it touches, and then every call site of `_markAggregateUnavailable`.

**Verdict: FAIL** (the round 0 BLOCKER is fixed at the door I named, and the same desync still reaches a client through a second door)

### Findings

- `src/SoilMoistureSystem.lua:2736-2746` (`settleDaily`): **BLOCKER**: pure-client MP desync, the same class as round 0. The fix gates `refreshForPublication` (`:1349`) and `_refreshFieldAggregate` (`:1400-1407`), but the daily settle's re-derive has no client gate. On a client whose streamed map has no pixels for a field, `settleDaily` marks it `UNAVAILABLE:EMPTY` or `INVALID_FIELD_GEOMETRY`. From then on the receive paths write the host's number, but the HUD, PDA and consultant show nothing.
  - **Bench proof:** on a scratch copy of c2d16c8 I added a probe row. With g_server nil, a live map with no pixels, and a field current at 0.6 and clean: S0 passes (the refresh gate keeps 0.6). After `settleDaily(1)`, S1 gets `nil:UNAVAILABLE`. After `CropStressMoistureInitEvent:run` delivers 0.55, S2 gets nil from `getFieldsSortedByMoisture`.
  - **Reachability:** I found no server gate on the path. `CropStressManager:update` runs the hourly block with no `g_server` test (`CropStressManager.lua:582-601`). It is held only by `isMissionWaterReady`. Every barrier input is set from ungated hooks (`main.lua:457` settings, `main.lua:465` load, `CropStressManager.lua:404-405` fields and map), and `onHourlyTick` calls `checkDayFallback` -> `settleDaily`. I have NOT proven in game that a client's barrier applies; that is the one link I could not close by reading.
  - **Fix:** apply the refresh rule at the settle too. On a client, OK with a mean marks the field current, PROVIDER_REFUSAL only fails closed, and every other outcome leaves the state alone. Add a pure-client settle row in the C shape, with a reached control.
- `src/HUDOverlay.lua:300-301`: **MAJOR** (the round 0 cost MAJOR, now larger): Fred's correction is right. `dt` is milliseconds (`mission.time = mission.time + dt`, `BaseMission.lua:770`), so `rebuildTimer >= 1.0` passes every frame, and the HUD calls `getFieldsSortedByMoisture` -> `refreshForPublication` every frame. On the host that is about one native `executeGet` per blank field per frame (~60 per second each), where round 0 estimated 2. The throttle is the same at 4a2b2f3, so the old per-frame sort is not new, but the per-frame native read is. It is named in the PR body, and Desk has ruled no behaviour change, so it does not block; I am only raising it again because the figure grew thirtyfold.

### Round 0 items

- **BLOCKER, refresh door:** fixed. The gate is at `:1349`. On a client, `_refreshFieldAggregate` keeps OK -> current and clean, keeps the refusal fail-closed without blanking, and leaves every other outcome alone (`:1400-1407`). Rows C1-C4 name it, with a reached control.
- **MAJOR 1:** fixed. Call site b now fires only on `outcome == "EMPTY"`, at `:1414` and `:2746`. H8c-f and D4b-c each have an EMPTY reached control on the same fixture, so they are not early exits.
- **MAJOR 2:** named in the PR body (see above).
- **MAJOR 3:** the sentence is in the PR body and TESTING row 49.
- **MINOR:** fixed. The ZONE cell with no seed keeps the water in field-wide `mapPending` (`:1816-1823`), and W4b names it.

### Other `_markAggregateUnavailable` sites checked on the client path

`SoilMoistureGround.lua:178/186/207` are behind the check's `g_server` guard. `:697` (pre-barrier migrate with no base) needs a field with no number and no last-known value, and a client's enumerated slot is numeric. `:773` (`_markRestoredFields`) needs `valueMap.loadedFromSave` on a client. I did not prove either of those two unreachable, only unlikely.

### Bench

`node run-tests.mjs` at c2d16c8 in the worktree: **2861 passed, 0 failed across 60 files**. The worktree was left clean, and the probe ran on a scratch copy.

**Not checked:** an in-game client join (see reachability above), and a re-run of Fred's 37 mutations.

### Commits on the branch (8)

```
c2d16c8 fix(RSF-F245/F247): review round 1: client keeps the host's number; call site b only on EMPTY
61ea779 test(RSF-F247): parcel probe count on every ground-check row; no-carrier migration never paints
7b28b0f test(RSF-F245/F247): pixel-grid bench for the moisture honesty pair
dd22531 fix(RSF-F245): a provider with no typed read is a capability refusal
d539446 feat(RSF-F245): host consumer floor never compares, ranks, averages, formats or sends a missing moisture
bf2535a feat(RSF-F245/F247): snapshot refresh, debug driest list, member-anchor irrigation calls
51de721 feat(RSF-F245/F247): restore, capture and carrier record keep absence absent
b3d1eea feat(RSF-F245/F247): aggregate honesty core, parcel probe and fill, field ground check, geometry retry doors
```

The title still matches all eight commits.

Bob

TheCodingDad-TisonK:
## Bob: re-ruling on my round 1 BLOCKER, PR #192 (c2d16c8)

**Head checked:** `refs/pull/192/head` = `c2d16c8`, unchanged.

**The round 1 BLOCKER is withdrawn.** Desk's reading is correct, and I checked it against c2d16c8 myself.
- `settleDaily` has exactly two callers (`git grep` at c2d16c8).
- `SoilMoistureSystem.lua:3055`, the Time Guard `onSettle` closure. It is registered only through `registerDailyAccrual`, whose single call at `CropStressManager.lua:390-391` sits behind `if g_server ~= nil`.
- `SoilMoistureSystem.lua:3076`, `checkDayFallback`. Its single call at `CropStressManager.lua:770` sits inside the `if g_server ~= nil then` block that opens at `:674` in `onHourlyTick` and closes after the broadcast.
- My chain was right up to `onHourlyTick`, and then I skipped the `:674` gate inside it. The probe row called `settleDaily` directly with g_server nil, a state no shipped caller produces. The bench result stands as a fact about the function, not as a reachable defect.

**Defensive guard in `settleDaily`: MINOR, not required.** Design's a2ae501 item 1 asked for explicit guards at Door C and `_checkFieldGround` specifically. The settle is not a door and is not public, and both its callers are gated. A one-line `g_server` rule at `:2736-2746` matching `_refreshFieldAggregate` would make a future ungated caller harmless. That is Fred's call: it is not a condition of this PR, and holding a merge for it would not be worth it.

**Verdict at c2d16c8: PASS with MAJORs.**
- Every round 0 finding is fixed and named (C1-C4, H8, D4, W4), and the suite is 2861/0.
- The one MAJOR left is the publication-door cost. It is named in the PR body, and Desk has ruled no behaviour change for this PR. The per-frame figure (dt in milliseconds, `BaseMission.lua:770`; about 60 native reads per second per blank field on the host) is with Tyson as a separate HUD throttle fix.

**Not checked:** an in-game client join, and a re-run of Fred's 37 mutations.

Bob

PR #193, opened 2026-09-17
SCS #191: csSimulateHeat never runs the hourly simulation on a client

### LINKED ISSUE #191: consoleSimulateHeat has no g_server gate, MP client can run hourly update locally
## What's wrong

`CropStressManager:consoleSimulateHeat` (`src/CropStressManager.lua:1729`) has no `g_server` gate. A player with console access (an admin on a multiplayer client, or anyone in singleplayer with the dev console on) can run `csSimulateHeat <days>` and it will call `self.soilSystem:hourlyUpdate(...)` and `self.stressModifier:hourlyUpdate()` directly, up to 24 times per day for up to 30 days.

On a multiplayer client this runs the hourly moisture/stress update **locally**, diverging that client's own field moisture and stress state from the server's authoritative state, since moisture is meant to be server-authoritative and synced to clients (per the mod's own README/CLAUDE.md: "Moisture sim is server-authoritative, clients receive synced state only").

## Where it was found

Found by Bob during the RSF-F245/F247 intake (ecosystem-dev-tracking ledger 35d585e). Pre-existing defect, not introduced by F245/F247.

## Scope

This is a console-command-only defect: it requires developer console access to trigger, so it is not reachable through normal play. Fix is planned as its own small PR, after F245/F247 lands.

## Suggested fix

Gate the function on `g_server ~= nil` (or `self:isServer()`, matching the pattern already used elsewhere in `CropStressManager.lua`, e.g. lines 390, 552, 568, 1062), the same way other server-authoritative paths in this file already are.

TheCodingDad-TisonK:
## SCS #191: csSimulateHeat never runs the hourly simulation on a client

Closes #191. Found by Bob during the RSF-F245/F247 intake (ledger 35d585e). Design (Iris, ledger a2ae501) named it a separate pre-existing defect, not absorbed into F245/F247.

### The defect
`CropStressManager:consoleSimulateHeat` had no server gate. On a multiplayer client it ran 24 soil and 24 stress hourly updates per simulated day on the client's own copy, so that client's moisture and stress diverged from the host. Moisture and stress are meant to be server-authoritative. There were two ways in:
- the `csSimulateHeat` console command (main.lua:332);
- the settings panel's **Simulate Heat Wave** admin button (CropStressSettingsPanel.lua:941), which a multiplayer client admin can click. That answers the open question in Bob's intake: the panel action is not host-only.

### The fix
- `consoleSimulateHeat` returns `false` and prints "csSimulateHeat runs on the host only" unless `g_server` exists. It uses the same `g_server ~= nil` test the manager already uses at :390, :552, :568 and :1068.
  - **Verified in the decompile:** `g_server = Server.new()` happens only in `MPLoadingScreen:startLocal` (singleplayer, :416) and `:startServer` (hosted and dedicated, :436). A joining client gets `g_client` only. `BaseMission` clears it on teardown (:116).
- The panel button reads that answer. On a client it now says "Heat wave simulation runs on the host only." instead of reporting a heat wave that never ran. The success message loses its em dash. No new l10n keys: both strings sit beside the panel's existing hardcoded English messages.

### Not changed
A client admin's click is not forwarded to the host. Doing that needs a new network event, which is a design question, not this fix.

### Tests
- New `tools/test/lua/SCS-191-console_heat_server_gate_spec_test.lua`, 14 assertions. It counts calls to the REAL `SoilMoistureSystem:hourlyUpdate`:
  - **Client:** the command returns false and runs no soil or stress hourly update; field moisture and the weather state are untouched. The panel button runs none either and shows the host-only message.
  - **Host:** one day runs 24 soil and 24 stress updates, the field dries and the weather is restored. The panel button runs 72 and reports the simulation.
- Suite: **2385 passed, 0 failed across 56 files** (base 2371/0/55).
- **Mutations, 3 of 3 killed by named rows:** gate removed (C1a-C2b), gate inverted (C and H rows), panel ignoring the answer (C2b). Every anchor matched once, the landed edit was proven, the full suite ran and each restore was sha256-checked.
- The Lua 5.1 syntax gate was proven on a broken control file first.

### Deploy
`bash build.sh --deploy` was not used, for the reason given on #192: with no zip binary it deletes the installed zip and then fails its copy. I ran `bash build.sh`, checked the listing (171 entries; modDesc, main and both changed files byte-identical to the worktree; no `tools/`; forward slashes) and copied the zip in.
- Installed sha256 **before** `e69104f3461db198ee34d0013e3f29da4ab0a64a94dfb6b23620886bbc4d5f1a` (the #192 build)
- **after** `ab578a35e293bcfad45df33630b1472041e3129bc0eb8710e4baf7ced1c04085` (this build)

The installed mod is now this branch, without the #192 changes.

### Not checked
No in-game session. log.txt is from 2026-09-15. The TESTING.md row carries the host, client and dedicated-server checks.


sasha-rf (approved):

Small, scoped fix for #191. g_server==nil gate matches engine behavior (g_server only set on singleplayer/host per MPLoadingScreen:startLocal/:startServer). Settings panel correctly branches on the false return with the client-facing message; the old success string's em dash is gone. Dedicated spec test included. Approved, awaiting Bob's cold diff verdict before merge-ready.

UTF-16 scan: 3 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #193 (SCS #191, csSimulateHeat server gate)

**Head checked:** `git ls-remote` `refs/pull/193/head` = `gh` headRefOid = `587118a`. Base `development` `4a2b2f3`. Full diff read (3 files, one commit).

**Verdict: PASS with MAJORs** (no BLOCKER, one MAJOR, one MINOR)

### Findings

- `ecosystem-dev-tracking/TESTING.md` row 50: **MAJOR** (doc): the row says that on the host "the client sees the drier values after the next sync". `consoleSimulateHeat` (`src/CropStressManager.lua:1736`) never calls `CropStressNetworkSyncBridge.markFieldDirty()` or broadcasts `CropStressMoistureInitEvent`. The only push is the hourly site (`CropStressManager.lua:796-803`), and NetworkSync reserializes only a module marked dirty (`CropStressNetworkSyncBridge.lua:245`). A client therefore sees the heat wave at the next in-game hour, not within a second, and a tester watching for a quick sync will record a false FAIL. That delay was already there before this PR and is not a regression. **Fix:** reword the row to "after the next in-game hour". Pushing the change immediately from the host path would be a separate item and is not needed for this fix.
- `src/settings/CropStressSettingsPanel.lua:941`: **MINOR**: the panel tests `== false`, while the usage branch of `consoleSimulateHeat` (`:1747`) returns nil. That branch is unreachable from the panel, which always passes "3", so there is nothing to fix now. If a later caller passes user input, use an explicit `true` on success and test `~= true`.

### Checks

- **Gate:** `if g_server == nil then return false end` runs before any weather override or hourly update, so a pure client neither mutates nor restores anything. `g_server = Server.new()` is created only at `gui/MPLoadingScreen.lua:416` (startLocal) and `:436` (startServer) in the decompile, so singleplayer, a hosted game and a dedicated server all keep the command.
- **MP:** moisture and stress stay server-owned, and a client admin gets an honest refusal. Forwarding a client click to the host would need a new event; that is design and correctly stays out.
- **Strings:** the two new English popup strings match the panel, which has no `g_i18n` use at all (count 0), so they are not drift. The em dash in the old success string became a colon.
- **Bench:** `node run-tests.mjs` at 587118a: **2385 passed, 0 failed across 56 files**; `SCS-191-console_heat_server_gate_spec_test.lua` 14/0. C1b/C1c count the real `SoilMoistureSystem:hourlyUpdate` wrapped, not a setup count. H1b (24 per day) and H2a (72 from the panel) prove the host fixture reaches the loop, so the client zeros are not an early exit. The worktree was left clean.
- **TESTING row:** present, with concrete singleplayer, hosted and dedicated steps (apart from the MAJOR above).

**Not checked:** any in-game run. I did not re-run Fred's three mutations. I also have not verified what the engine does with a `false` returned from an `addConsoleCommand` callback (`main.lua:332`); it is not in the references.

### Commits on the branch (1)

```
587118a fix(SCS #191): csSimulateHeat never runs the hourly simulation on a client
```

The title "SCS #191: csSimulateHeat never runs the hourly simulation on a client" matches the commit.

Bob

PR #194, opened 2026-09-17
chore(build): --deploy never removes the installed zip before the new one is good

TheCodingDad-TisonK:
### The defect
`bash build.sh --deploy` deleted the installed `FS25_SeasonalCropStress.zip` first, then copied the new zip over. When the copy source did not exist, the mod was gone and the copy failed, so nothing was installed. That happens on every build from a git worktree on a machine without a `zip` binary: the Python fallback names the zip after the folder, but the copy looks for `../FS25_SeasonalCropStress.zip`.

### The change (deploy block only)
- **Built zip checked first.** The zip this run built must exist and be non-empty. When `unzip` is available, it must also list `modDesc.xml` at its root. The script already removes any old zip at `ZIP_PATH` before building, so an existing zip is this run's.
- **Staged copy.** It is copied beside the installed zip as `FS25_SeasonalCropStress.zip.deploying` and compared byte for byte (`cmp`).
- **Replace.** Only then does `mv -f` move it over the installed zip.
- **On failure.** Every failure before that move removes the staged copy, prints "The installed mod was NOT touched." and exits 1. If the move itself fails, for example because the game holds the zip, the installed zip stays as it was.

### Proof: sandbox
`USERPROFILE` points at a fake home, so `MODS_DIR` is a fake mods folder holding a fake installed zip. Hashes are the first 16 hex characters of sha256; each test starts from a fresh fake zip. The real installed zip was never touched: `7c46c31c...` before and after the whole run.

| Test | Script | Case | build.sh exit | Installed zip |
|---|---|---|---|---|
| T1 | old (4a2b2f3) | --deploy from a worktree (zip-name mismatch) | 1 | `84a0e9de...` -> **ABSENT** (deleted) |
| T2 | new | same worktree case | 1 | `17da534e...` -> `17da534e...` ("no built zip", not touched) |
| T3 | new | good zip, staging copy cannot be made (a directory sits at the `.deploying` name) | 1 | `634d45ef...` -> `634d45ef...` |
| T4 | new | build fails (no zip and no Python on PATH) | 1 | `2f0e6b07...` -> `2f0e6b07...` |
| T5 | new | built zip has no modDesc.xml | 1 | `8e5b6119...` -> `8e5b6119...` |
| T6 | new | good zip, normal deploy | 0 | `2d47cf9c...` -> `43834c4a...`, equal to the built zip; no `.deploying` left; `modDesc.xml` and `main.lua` at root |

### Proof: real mods folder
`bash build.sh --deploy` from this branch's worktree (`_wt-scs-deploy`), the exact case that used to delete the mod:
- exit 1: "ERROR: no built zip at .../FS25_SeasonalCropStress.zip", "The installed mod was NOT touched."
- Installed sha256 **before** `7c46c31c5a827708d8f2b627993ff667eaa1442699ef746ab537401a6198c4a2`, **after** the same. No `.deploying` file left.
- The zip the fallback did build (`../_wt-scs-deploy.zip`, sha256 `788b39f8...`) lists 171 entries. modDesc.xml and main.lua are at root and byte-identical to the worktree, with no `tools/` and forward slashes. `build.sh` is not in the zip, so this change ships nothing to players.

### Still open, separate and untracked (not fixed here)
1. **Zip name.** The Python fallback names the zip after the folder (`os.path.basename(MOD_DIR) + ".zip"`), while the deploy step copies `../FS25_SeasonalCropStress.zip`. From a worktree, `--deploy` now fails safely instead of deleting the mod, but it still cannot deploy.
2. **.git pointer.** A worktree's `.git` pointer file is still packed into the zip (the Python fallback excludes `.git` only as a directory; the `zip` branch excludes only `./.git/*`). The zip above still contains it.

### Not checked
- The `zip`-binary branch of the builder: this machine has no `zip`. The deploy block is shared, so it runs the same checks.
- A deploy while the game holds the installed zip open.
- A machine without `unzip`, where only the size check runs; the script prints a line saying so.


sasha-rf (approved):

Build-script-only, no player-visible change (build.sh ships in no zip). Deploy now stages a byte-verified copy and does an atomic mv over the installed zip instead of rm-then-cp, so a failure anywhere before the move leaves the installed mod untouched. Checks non-empty zip and modDesc.xml presence first. Sound fix for the same class of deploy-safety gap named in the SCS #192 and MD #155 PRs. Approved, awaiting Bob.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #194 (SCS build.sh safe deploy)

**Head checked:** `git ls-remote` `refs/pull/194/head` = `gh` headRefOid = `e24030a`. Base `development`. One file, the deploy block only.

**Verdict: PASS**

### Findings

- `build.sh` (the `.deploying` stage): **MINOR**: if the run is killed between the `cp` and the `mv` (Ctrl+C, a closed terminal), `FS25_SeasonalCropStress.zip.deploying` stays in the mods folder until the next deploy's `rm -f` clears it. The game loads only `.zip`, so it is harmless. No fix is needed; a `trap` would be optional.

### The three asks

1. **set -e:** safe. Each failing step sits in an `if` / `if !` condition (`[ ! -s ]`, `command -v`, `unzip -l | grep -q`, `cp || cmp`, `mv -f`), and bash does not apply `-e` there. The cleanups are `rm -f ... || true`. There is no `pipefail`, so the unzip pipeline's status is grep's, which is the intended test. A `grep -q` early exit that SIGPIPEs unzip cannot turn a match into a failure.
2. **`mv -f` over a held zip on Windows: tested, and it never leaves neither file.** I held `dest.zip` open from PowerShell (`[IO.File]::Open`, read access) and ran Git's `/usr/bin/mv -f dest.zip.deploying dest.zip`:
   - share `Read`: exit 1, "Device or resource busy", dest still OLD, staged left (the script then removes it)
   - share `ReadWrite`: the same as Read
   - share `ReadWrite, Delete`: exit 0, dest is NEW, no staged file left

   So the only outcomes are old-kept-and-refused or new-in-place. I did not observe which share mode FS25 itself uses on an open mod zip.
3. **The grep anchor `" modDesc.xml$"` against real `unzip -l` output:** sample zips checked. Root `modDesc.xml` matches. `xml/modDesc.xml`, `FS25_X/modDesc.xml` and `oldmodDesc.xml` do not.

### Checks

- The old flow deleted the installed zip before the copy could fail. In the new flow nothing touches `DEST` until a non-empty zip listing a root `modDesc.xml` has been staged beside it and `cmp`-verified.
- `ZIP_PATH` is removed at the top of the script and a failed build exits under `set -e`, so the deploy block cannot pick up a stale zip from this path. The worktree zip-name mismatch (the Python fallback names the zip after the folder) is caught by `-s` and named in the PR as a separate defect.
- **TESTING row:** row 52 exists and names concrete steps.

**Not checked:** a real build or deploy (Bob does not build), Fred's T1-T6 sandbox runs (I did not re-run them), and the game actually holding the zip.

### Commits on the branch (1)

```
e24030a chore(build): --deploy never removes the installed zip before the new one is good
```

The title matches the commit.

Bob

PR #195, opened 2026-09-17
fix(HUD): the row-rebuild throttle counts milliseconds, once per 1000 ms

TheCodingDad-TisonK:
### The defect
Line numbers below are at base 8d7dbd59; this PR adds two lines at :45, so later lines move down by 2.

`HUDOverlay:update(dt)` receives `dt` in milliseconds: `BaseMission:update` adds it to mission time (BaseMission.lua:770) and passes the same value to mod listeners (:796-798). The row-rebuild throttle compared its accumulator to `1.0` (HUDOverlay.lua:300-301), so while the HUD is visible the list rebuilt on every frame instead of once a second. Since RSF-F245 (#192), each rebuild runs `getFieldsSortedByMoisture` -> `refreshForPublication`, so every frame costs one native polygon read per field with no current reading. That is the MAJOR on #192 that Tyson ruled on.

### The change (minimal)
- `HUDOverlay.REBUILD_INTERVAL_MS = 1000`, with a one-line comment on the units.
- The throttle compares `rebuildTimer >= HUDOverlay.REBUILD_INTERVAL_MS`.
- `onMoistureUpdated` sets `rebuildTimer = HUDOverlay.REBUILD_INTERVAL_MS`. Its force still rebuilds on the next tick; with a 1000 ms threshold, the old `1.0` would have stopped forcing anything.

### Proof
New `tools/test/lua/SCS-HUD-rebuild_throttle_spec_test.lua` drives the real `HUDOverlay:update` with a counted `rebuildDisplayRows`:
- **R1:** 49 frames of 20 ms (980 ms) give no rebuild. The 50th frame (1000 ms) gives one. **R1c:** 3000 ms of 20 ms frames give 3 rebuilds, not 150.
- **R2a:** after `onMoistureUpdated`, the next 20 ms tick rebuilds. **R2b:** the throttle then resumes.
- **On the unfixed code** (base 8d7dbd59) the spec fails 4 of 5: 49, 50 and 150 rebuilds. R2a passes there only because every frame rebuilt.
- **Mutations, both killed by named rows:**
  - H01 (throttle back to `1.0`): R1a-c, R2b.
  - H02 (force back to `1.0`): R2a, R2b.
  - The full suite ran each time, and each restore was sha256-checked.
- **Suite:** 2390 passed, 0 failed across 57 files.
- **Merge check with #192:** `git merge-tree` of this head against #192's head c2d16c82 is clean. #192 edits HUDOverlay.lua only in the row drawing, grouped sort, forecast and stub rows. This PR touches line 45, the throttle at :300-301 and the force at :1658.

### Deploy
`bash build.sh`, listing checked (171 entries; modDesc.xml, main.lua and HUDOverlay.lua byte-identical to the worktree), copied in. `--deploy` from a worktree still refuses safely (the untracked zip-name defect, queue item 4).
- Installed sha256 **before** `7c46c31c5a827708d8f2b627993ff667eaa1442699ef746ab537401a6198c4a2` (the #192 build)
- **after** `c1afb3d9111771aadb40f49186959828f3e45dca22ab117d63eca25f4e6bc56e` (this build, without #192)

### Other seconds-versus-milliseconds timers (untracked, not fixed here)
- **`autoHideTimer`** is declared "real-time seconds" (HUDOverlay.lua:158) and set to `120` by `onCriticalThreshold` (:1628), but `update` subtracts millisecond `dt` from it (:316). The HUD shown for a critical alert auto-hides after about 120 ms instead of 120 s.
- **`animTimer`** accumulates milliseconds (:257) and feeds `math.sin(self.animTimer * 4)` for the pulse at :927 and :998. That is 4 radians per millisecond, far above the frame rate, so the pulse samples as jitter rather than a smooth beat.
- **`_vehicleDebugTimer`** adds a fixed `0.016` per call and fires at `5.0` (:898-899). That counts frames, not time (about 5 s only at 60 fps), and it is debug only.

Every other `dt` timer in `src` already uses milliseconds: CsPDAScreen.lua:47 and :867, RfPdaMenuPage.lua:59-61, the vendored rainstarAutoReel (`ANALYSIS_INTERVAL_MS`, `completeMessageTimer = 5000`), and IrrigationManager's `/ 3600000`.

### Not checked
- **No in-game run.** The frame-time gain with the HUD open has not been measured.


sasha-rf (approved):

Matches the throttle bug named in #192's PR body exactly: dt is milliseconds (cited against BaseMission.lua:770/796-798), but the old compare used 1.0 instead of 1000, so the "once a second" throttle ran nearly every frame. Fix is a named constant (REBUILD_INTERVAL_MS = 1000) used at both the throttle compare and the force-rebuild-on-moisture-update site. Spec test covers cadence (R1a/b/c: no rebuild before 1000ms, one at 1000ms, three over 3000ms) and the force-then-resume path (R2a/b). Small, correct, well tested. Approved, awaiting Bob.

UTF-16 scan: 2 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #195 (HUD rebuild throttle in milliseconds)

**Head checked:** `git ls-remote` `refs/pull/195/head` = `gh` headRefOid = `e22913d`. Base `development`. Two files, one commit.

**Verdict: PASS**

### Findings

- `src/HUDOverlay.lua:160` and `:1630`: **MINOR** (older than this PR; named in the PR body): `autoHideTimer` is commented as seconds and set to 120, but `:318` counts it down by `dt` in milliseconds, so the auto-hide fires after 0.12 s rather than 2 minutes. `animTimer` feeds `sin(ms * 4)` at `:929` and `:1000`, a pulse far faster than intended. **Fix:** a separate item for each, not this PR.

### Checks

- **Unit:** `dt` is milliseconds. The chain is `main.lua:343` (appended `FSBaseMission.update(self, dt)`) -> `CropStressManager:update(dt)` (`:534`) -> `hudOverlay:update(dt)` (`:625`), and the engine adds that same `dt` to mission time (`BaseMission.lua:770`, `self.time = self.time + dt`).
- **Both sites changed:** the throttle (`:302-303`) and the `onMoistureUpdated` force (`:1660`) use `REBUILD_INTERVAL_MS`. The force value equals the threshold, so `>=` still fires on the next tick. No other file reads or writes `rebuildTimer` (`git grep` at e22913d).
- **Bench, on a `git archive` copy:** 2386 passed, 0 failed across 57 files (plus `scs018_decay_catchup`, which cannot load in a copy with no sibling `FS25_TimeGuard`; that is a copy artifact). `SCS-HUD-rebuild_throttle_spec_test.lua` 5/0.
- **Mutation H01 re-run (the throttle back to `>= 1.0`, one hunk confirmed):** **KILLED** by name. R1a got 49 want 0, R1b got 50 want 1, R1c NAMED got 150 want 3, and R2b fails as well. R1a is a reached control on the same fixture, so the counts come from the real `update()` path, not from setup.
- **#192:** `git merge-tree --write-tree e22913d c2d16c8` is clean. With this fix, #192's publication refresh on the HUD path drops from every frame to once per 1000 ms. That shrinks the #192 cost MAJOR back to about the round 0 estimate.
- **TESTING row:** present.

**Not checked:** an in-game run, and a re-run of H02 (Fred's force mutation).

### Commits on the branch (1)

```
e22913d fix(HUD): rebuild throttle counts milliseconds, once per 1000 ms
```

The title "the row-rebuild throttle counts milliseconds, once per 1000 ms" matches the commit (its wording differs slightly, same meaning).

Bob

PR #164, opened 2026-09-01
feat: SCS-046 pivot rain key engine (fitted pause, sensor, gate, snapshot)

TheCodingDad-TisonK:
SCS-046 center-pivot rain key, engine half.

An optional rain key per center pivot watches current rain at that machine. Meaningful rain trips the key and pauses irrigation water, pivot movement and operating cost together while the schedule stays intact; 30 continuous readable dry game minutes clear the latch without surprise-starting the pivot. Unfitted pivots are bit-for-bit unchanged.

What's in:
- Per-pivot rain-key state (fitted, trip mm, accumulated mm, dry minutes, latch, input state, revision) in IrrigationManager.
- WeatherGuard-first current-rain read (getCurrentSky under pcall, base-game fallback, UNAVAILABLE when both fail; nil is never zero or dry).
- Continuous sensor tick in CropStressManager:update before the hourly branch; effective rain is isRaining with rainScale >= 0.05; trip once; 30 dry game minute reset.
- One operational gate in activateSystem and hourlyScheduleCheck: a fitted tripped or input-unavailable row refuses every scheduled, remote and one-shot start.
- Fractional active-hour settle: fitted pivots settle water + cost through the existing footprint and never enter the legacy whole-hour gain or chargeHourlyCosts pass.
- Copy-only getIrrigationSystems snapshot expansion (rainKeyState, activityState, pauseReason, nextWakeKind, stateRevision).
- Server-authoritative FIT / REMOVE / SET_TRIP_MM command + result events.
- csRainKeyCheck diagnostic.
- Release-gate row rain_key_pause LOCKED.

47 assertions in SCS-046-pivot-rain-key_spec_test.lua. Not yet in-game verified. The FarmTablet readout is Wizard's half and travels with this.

PR #196, opened 2026-09-17
chore(build): keep a worktree's .git pointer file out of the zip

TheCodingDad-TisonK:
## chore(build): keep a worktree's .git pointer file out of the zip

This is the SCS counterpart of MarketDynamics #155 and FarmTablet #164, and one of the two worktree defects named on #194.

In a git worktree `.git` is a small file that points at the main clone's git directory, not a directory. `build.sh` has two packers, and both let that file through:
- **Python fallback** (used when there is no `zip` binary): `.git` was only in `EXCLUDE_DIRS`. It is now also in `EXCLUDE_FILES`, with a comment saying why.
- **`zip` branch:** only `--exclude "./.git/*"`. `--exclude "./.git"` is added.

Base development 8d7dbd59; the #194 deploy block is untouched.

### Proof
Both builds ran with `bash build.sh` from worktree `_wt-scs-gitptr`, through the Python fallback (this machine has no `zip`), before and after the change.

| Build | Entries | `.git` entry | sha256 |
|---|---|---|---|
| Before | 171 | `.git` (file) | `ae985329689d7ba908cda17eb0d94670b88c3e13cb3020de54f35a9db53d86c1` |
| After | 170 | none | `66872351a4d784e84fa8fb4041ae5f979ba6ec15102b8e8ab39fa66ccc733389` |

The sorted listings differ by exactly one line, `.git`.

**Installed zip.** `--deploy` from a worktree still refuses (the zip-name defect, queue item 4, a separate PR), so the built zip was copied in.
- sha256 **before** `c1afb3d9111771aadb40f49186959828f3e45dca22ab117d63eca25f4e6bc56e` (the #195 build)
- **after** `66872351a4d784e84fa8fb4041ae5f979ba6ec15102b8e8ab39fa66ccc733389`, equal to the built zip: 170 entries, no `.git`.

### Not checked
- **The `zip`-binary branch.** There is no `zip` here, so the new `--exclude "./.git"` is by reading. It uses the same `./`-relative form as the script's other patterns.
- **A main-clone build.** There `.git` is a directory and is still pruned, so that listing should not change.
- **An in-game load.** The mod files are unchanged; only the pointer entry is gone.


sasha-rf (approved):

Build-only, no player-visible change. Same .git-worktree-is-a-file defect as MD #155/FT #164, but build.sh has two code paths (native zip and the Python fallback); this fix covers both - the existing "./.git/*" exclude only matched a directory, so it adds a bare "./.git" exclude for the zip branch and .git to EXCLUDE_FILES for the Python fallback. Approved, awaiting Bob.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #196 (SCS build.sh .git pointer)

**Head checked:** `git ls-remote` `refs/pull/196/head` = `gh` headRefOid = `e070fa4`. Base `development`. `build.sh` only, one commit.

**Verdict: PASS**

### Findings

None blocking.

- **Not verified, zip-binary branch (`build.sh:40`):** `--exclude "./.git"` follows the script's own `./` pattern form, which the neighbouring `./.git/*`, `./*.sh` and `./.gitignore` excludes already rely on. There is no `zip` binary on this machine either, so I could not run it. Worth one real build on a machine that has Info-ZIP before anyone relies on that branch in a worktree.

### Checks

- **Python fallback, run independently:** I extracted `git archive e070fa4` into a scratch folder named `FS25_SeasonalCropStress` (so the fallback's folder-derived zip name matches), added a `.git` pointer file, and ran the heredoc Python from `origin/development` and from e070fa4. Old: 171 entries including `.git`. New: 170 entries, no `.git` anywhere. This matches Fred's listing.
- The fallback prunes directories by relative path and skips files by bare name. In a main clone `.git` is a directory and was already pruned, so nothing changes there.
- The #194 deploy block is untouched. **TESTING row:** present.

**Not checked:** the zip-binary branch (above), a main-clone build, the installed hash, an in-game load.

### Commits on the branch (1)

```
e070fa4 chore(build): keep a worktree's .git pointer file out of the zip
```

The title matches the commit.

Bob

PR #197, opened 2026-09-17
chore(build): the Python fallback names the zip after MOD_NAME

TheCodingDad-TisonK:
## chore(build): the Python fallback names the zip after MOD_NAME

The second worktree defect named on #194.

### The defect
When there is no `zip` binary, `build.sh` packs through a Python fallback. It named the zip after the folder (`os.path.basename(MOD_DIR) + ".zip"`), while the shell's `ZIP_PATH` and the `--deploy` step use `../FS25_SeasonalCropStress.zip`. From a git worktree the folder is not named after the mod, so the zip landed at `../_wt-<name>.zip` and deploy could not find it. Since #194 that refusal is safe (the installed mod is not touched), but a worktree deploy could never succeed.

### The change
- The fallback is invoked as `CS_MOD_NAME="$MOD_NAME" $PYTHON_CMD - <<'PYEOF'`.
- Python names the zip `(os.environ.get("CS_MOD_NAME") or os.path.basename(MOD_DIR)) + ".zip"`. The folder name remains only as a fallback, with a two-line comment saying why.
- In the main clone the name is the same as before.

### Proof (real mods folder, worktree `_wt-scs-zipname`)
| Build | `bash build.sh --deploy` | Zip written | Installed sha256 |
|---|---|---|---|
| Base 8d7dbd59 | exit 1: "no built zip at .../FS25_SeasonalCropStress.zip", "The installed mod was NOT touched." | `../_wt-scs-zipname.zip` | `66872351...` -> `66872351...` (unchanged) |
| This branch | exit 0: "Deployed" | `../FS25_SeasonalCropStress.zip` | `66872351...` -> `a3636298adb0f62c85303760318de1c59ebe9d0d54d6c00f2d8e086bf07e6a22` |

After the fix:
- The built zip's sha256 is `a3636298...`, equal to the installed zip.
- No `.deploying` file is left behind.
- 171 entries; modDesc.xml, main.lua and HUDOverlay.lua are byte-identical to the worktree.
- This branch is off development without #196, so this zip still carries the `.git` pointer. #196 removes it.

### With #196
Both PRs edit build.sh at different lines. `git merge-tree` of this head (854bb72) with #196's head (e070fa4) is clean, so whichever merges second should not need a rebase.

### Not checked
- **The `zip`-binary branch.** It already names the zip `$ZIP_PATH`, and this machine has no `zip`.
- **Two builds at once.** Main-clone and worktree builds now share `../FS25_SeasonalCropStress.zip`, so two concurrent builds from sibling folders would overwrite each other's zip. The main clone always used that path.
- **An in-game load.**


sasha-rf (approved):

Build-script-only, no player-visible effect. Fixes exactly the zip-name mismatch named (not fixed) in #192's PR body: the Python fallback used to name the zip after the working directory's basename, which is wrong in a git worktree (e.g. a folder named for the branch, not the mod). MOD_NAME is threaded through via CS_MOD_NAME env var into the heredoc and used first, with the folder basename kept only as fallback. Confirmed MOD_NAME is set at build.sh:11 before this. Approved, awaiting Bob.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #197 (SCS build.sh worktree zip name)

**Head checked:** `git ls-remote` `refs/pull/197/head` = `gh` headRefOid = `854bb72`. Base `development`. `build.sh` only, one commit.

**Verdict: PASS**

### Findings

- `build.sh:14` / `:62` (named in the PR, not fixed): **MINOR**: the main clone and every SCS worktree now write the same `../FS25_SeasonalCropStress.zip`, and each build's first step deletes it (`:24-25`). Two concurrent builds, for example two seats, can remove or overwrite each other's zip mid-build. #194's `-s` and modDesc checks would refuse a missing zip, but not one written by the other build. **Fix:** a separate chore if seats ever build SCS in parallel; not needed here.

### Checks

- **Same path as the shell:** Python's `dirname(MOD_DIR)` with `MOD_DIR = os.getcwd()` after `cd "$SCRIPT_DIR"` (`:32`) is the shell's `OUTPUT_DIR` (`$SCRIPT_DIR/..`), and the name is now `MOD_NAME`, so `ZIP_PATH` agrees in both.
- **Env prefix:** `CS_MOD_NAME="$MOD_NAME" $PYTHON_CMD - <<'PYEOF'` is a valid per-command environment assignment in bash. The folder-name fallback remains when the variable is absent.
- **Independent run:** I extracted `git archive 854bb72` into a scratch folder named `_wt-anything` and ran the heredoc Python. With `CS_MOD_NAME` set it wrote `FS25_SeasonalCropStress.zip`; without it, `_wt-anything.zip`. Both branches behave as written.
- **With #196:** `git merge-tree --write-tree 854bb72 e070fa4` is clean. The #194 deploy block is untouched. **TESTING row:** present.

**Not checked:** a real `--deploy` (Bob does not deploy), the zip-binary branch (unchanged here), an in-game load.

### Commits on the branch (1)

```
854bb72 chore(build): the Python fallback names the zip after MOD_NAME
```

The title matches the commit.

Bob

PR #198, opened 2026-09-17
fix(HUD): critical-alert auto-hide and edit pulse count milliseconds

TheCodingDad-TisonK:
## fix(HUD): critical-alert auto-hide and edit pulse count milliseconds

These are the two seconds-versus-milliseconds timers named as untracked on #195, now assigned by Tyson. Base development ee06fe6d (includes #192 and #195).

### The defects
`HUDOverlay:update(dt)` receives milliseconds (BaseMission.lua:770 adds `dt` to mission time; #195 fixed the row-rebuild throttle for the same reason). Line numbers are at ee06fe6d.
- **Auto-hide.** `onCriticalThreshold` sets `autoHideTimer = 120` (:1651), declared "real-time seconds" (:160). `update` counts it down by `dt` (:317-322). The HUD a critical alert pops up hides itself after about 120 ms instead of two minutes.
- **Pulse.** The edit-mode border pulse is `0.5 + 0.5 * math.sin(self.animTimer * 4)` (:929, :1000), and `animTimer` accumulates `dt` (:259). That is 4 radians per millisecond: at 16 ms frames each frame advances 64 radians, so the pulse samples as jitter instead of a beat.

### How I read the intended pulse rate
The pulse came in with 044d841, "upgrade HUDOverlay to NPCFavor drag/resize/edit pattern", copied from NPCFavor's HUD (`NPCFavorHUD.lua:727`, `math.sin(self.animTimer * 4)`). NPCFavor accumulates the same millisecond `dt` (:596), so it carries the same bug and says nothing about intent on its own.

SoilFertilizer's SoilHUD was built from the same NPCFavor edit pattern (afd73446, "redesign SoilHUD with drag/resize edit mode (NPCFavor pattern)"). It writes that pulse as `math.sin(self.animTimer * 0.004)` on a millisecond `animTimer` (SoilHUD.lua:829, :1629). That is the same coefficient 4, applied per second: 4 radians per second, a period of 2*pi/4, about 1.571 s. This PR uses that rate.

### The change
- `HUDOverlay.AUTO_HIDE_MS = 120000`, set by `onCriticalThreshold`. The field comment now says "milliseconds left".
- `HUDOverlay.PULSE_RAD_PER_MS = 0.004`, used through one new `HUDOverlay:editPulse()` helper by both draw sites.
- The countdown and the accumulator themselves are unchanged.

### Proof
[...]

PR #199, opened 2026-09-17
feat(SCS #191): a client admin's heat wave runs on the host, after a server rights check

TheCodingDad-TisonK:
## feat(SCS #191): a client admin's heat wave runs on the host, after a server rights check

Design ruling d3c0626 (Arissani, recorded by ClaudeA), built to Bob's intake 82af751 (FAST TRACK) and his correction to it (use `run` from `readStream`, not the RainKey shape). Base development ee06fe6d.

#193 refused Simulate Heat Wave on every pure client. The ruling: a client admin's console command or panel click sends a request, the server re-checks admin rights itself and runs the existing host simulation, #193's refusal stays for a non-admin or rejected request, and the player always sees the truth.

### The request and its answer
- **`CropStressHeatRequestEvent`** (client -> server, `UInt8 days`). In `run`, the server:
  - **checks rights itself.** The sending connection must not be a server connection, and must be local or a master user: `connection:getIsLocal() or g_currentMission.userManager:getIsConnectionMasterUser(connection)`. That is the engine's test for server admin actions (UserManager.lua:165-169, SaveEvent.lua:13, CheatMoneyEvent.lua:18). Farm permissions are not used.
  - **validates days** to a whole number 1-30 and never trusts the stream (`CropStressManager.validHeatDays`).
  - on accept, runs `CropStressManager:runHeatSimulation(days)`, which is the old `consoleSimulateHeat` body plus a push to clients.
  - **answers only the requesting connection** with `CropStressHeatResultEvent` (`Bool accepted, String code, UInt8 days`; codes `OK`, `NOT_ADMIN`, `BAD_DAYS`).
- **Both events call `self:run(connection)` at the end of `readStream`.** The engine only calls `readStream` on an incoming event and never `run` (Server.lua:435-437; `run` is called directly only for a local-stream send, Connection.lua:73-74). The RainKey pair the intake first pointed to lacks this. That is named below as a separate defect, not copied.
- **Sourced** in main.lua beside the other events; `InitEventClass` is at file scope.

### One entry, no early success
- `CropStressManager:requestHeatWave(days, origin)` is the single entry for the `csSimulateHeat` console command and the panel's Simulate Heat Wave button.
  - **Host or singleplayer:** runs `runHeatSimulation` directly and reports it.
  - **Pure client:** sends the request and shows only "Heat wave request sent to the server."
  - **No server connection:** "Heat wave request could not be sent to the server."
- `CropStressManager:onHeatResult` shows the server's answer where the request came from: the panel as a new popup, the console as a printed line.
- **Refusal wording** is now "Only a server admin can run the heat wave simulation." #193's "runs on the host only" is no longer true for an admin.

### Hourly-push delay (in scope per the intake)
The hourly act's client push is now `CropStressManager:pushMoistureToClients()`: NetworkSync `markFieldDirty` when active, else refresh and broadcast `CropStressMoistureInitEvent`, unchanged. `runHeatSimulation` calls the same helper, so a client admin sees the heat wave at once instead of at the next in-game hour.

### Pre-existing defect fixed here (per the intake)
`CropStressSettingsPanel:isAdmin()` called `user:getIsAdmin()`, which does not exist: `users/User.lua` defines `getIsMasterUser` (:93) and no `getIsAdmin`, and it is not in any reference. It threw in multiplayer whenever the user lookup succeeded. It now returns `g_currentMission.isMasterUser == true` (set for this player in `FSBaseMission:onMasterUserAdded`, :2974-2976). That value is for display only; the server re-checks every request.

### Proof
New `tools/test/lua/SCS-191-heat_admin_request_spec_test.lua`, 29 assertions. **Every event is delivered through a stream round trip** (`writeStream` -> a fresh instance's `readStream`), never by calling `run` directly.
- **S, server receipt:**
  - A master user's request for 2 days runs 48 soil and 48 stress hourly updates, replies `true:OK:2` to that connection, and pushes to clients once (S1). The stream reads cleanly (S1a).
  - A non-master runs nothing, gets `NOT_ADMIN` and pushes nothing (S2). A local connection is accepted (S3). A request over a server connection is refused (S4).
  - Forged day counts 0, 31 and 255 run nothing and get `BAD_DAYS` (S5). A non-master with a forged count still gets `NOT_ADMIN` (S6).
- **C, a pure client:**
  - The console sends exactly one request for 3 days and prints only "request sent" (C1). The panel sends exactly one request and pops up only "request sent" (C2).
  - The server's accepted answer, off the stream, pops up the simulated text (C3). A `NOT_ADMIN` answer pops up the refusal (C4). A console request's answer prints (C5).
  - With no server connection nothing is sent and the player is told (C6).
- **E, end to end:** a client admin's click produces a request, the server runs 72 updates, the reply produces the simulated popup.
- **A, isAdmin():** in multiplayer it reads `isMasterUser` with no error: true for the master user, false otherwise.
- **The #191 spec** (`SCS-191-console_heat_server_gate_spec_test.lua`) rows move to the new truth. A client's console still never simulates locally (C1a); the client panel never claims a simulation ran (C2b). The host rows expect `"RAN"` and the new success text. The fixture gains a `broadcastEvent` on `g_server` for the push.
- **Mutations, all killed by named rows** (full suite each, each restore sha256-checked):
  - Q01 rights check removed: S2a-c, S4, S6.
  - Q02 push removed: S1e.
  - Q03 success claimed before the reply: C1c, C2b.
  - Q04 request `readStream` without `run`: S1b-e, E1a.
  - Q05 result `readStream` without `run`: C3, C4, C5, E1b.
  - Q06 day validation removed: S5 (a forged 255 ran 6120 updates).
- **Suite:** 2909 passed, 0 failed across 63 files. The Lua 5.1 syntax gate passes.

### Deploy
`bash build.sh --deploy` from the worktree: 173 entries, both new event files and the three changed files byte-identical to the worktree.
- Installed sha256 **before** `d2fdfa2d0e6ae9e29f29dc29c2c259f5f40e18ce67de3359fa81ace9df526649` (the #198 build)
- **after** `4ea89f0b601c117fded4b899ec355625c3548e85a2c71a49b3fdccb9a8aa71cc`, equal to the built zip.
- This branch is off ee06fe6d, so the installed mod no longer carries #198's HUD timer fix until both merge.

### Separate, untracked (found while building this, not fixed here)
Five existing SCS events end `readStream` without calling `run`:
- `CropStressIrrigateNowEvent`
- `CropStressPivotRemoteEvent`
- `CropStressScheduleSyncEvent`
- `CropStressRainKeyCommandEvent`
- `CropStressRainKeyResultEvent`

By reading the engine, a pure client's Irrigate Now, pivot remote, schedule save and rain-key commands are received by the server and discarded. Sent to Desk; Bob confirmed Server.lua:436-437.

### Not checked
- **No in-game multiplayer run:** a client admin's click on a listen host or a dedicated server, and a non-admin's refusal, have not been played.
- **Connection and user manager are mocks** of the decompiled methods, not the engine.


sasha-rf (approved):

Matches Design's ruling (d3c0626) exactly: a new CropStressHeatRequestEvent carries only days over the wire, and the server re-checks admin rights itself at receipt via UserManager:getIsConnectionMasterUser (the same pattern SaveEvent/CheatMoneyEvent use, cited) - never trusts the client's claim. CropStressHeatResultEvent answers only the requesting connection with accepted/code/days, and requestHeatWave() cleanly unifies the host-runs-directly and client-sends-and-waits paths for both the console command and the settings panel. Separately fixes a real Lua error: CropStressSettingsPanel:isAdmin() called user:getIsAdmin(), which doesn't exist on User (decompiled User.lua:93 has getIsMasterUser); replaced with the mission's own isMasterUser flag (set in FSBaseMission:onMasterUserAdded, cited with line numbers), display-only since the server re-checks anyway. Test coverage uses a real writeStream/readStream round trip through a mock stream, not a direct function call, which is the right level for network code. Approved, awaiting Bob.

UTF-16 scan: 7 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #199 (SCS #191 client-admin heat wave request)

**Head checked:** `git ls-remote` `refs/pull/199/head` = `gh` headRefOid = `1c6d00f`. Base `development` ee06fe6d. Seven files, one commit. Built to intake 82af751 and its correction d987b41.

**Verdict: PASS with MAJORs** (no BLOCKER, one MAJOR, one MINOR)

### Findings

- `src/CropStressManager.lua:795` (`onHourlyTick`): **MAJOR** (missing test): the hourly push moved into `pushMoistureToClients()`, but nothing pins that the hourly site still calls it. My mutation replaced `self:pushMoistureToClients()` at the hourly site with a comment (one hunk confirmed). **SURVIVED**: the suite was 2905/0 and the heat spec 29/0. A later edit could silently drop every hourly client update. **Fix:** a named row that runs `onHourlyTick` on a server with the bridge inactive and asserts exactly one `CropStressMoistureInitEvent` broadcast, plus a reached control that proves the hourly body ran. A second row with the bridge active should assert `markFieldDirty` is called once.
- `src/CropStressManager.lua` (`consoleSimulateHeat`): **MINOR**: it prints `text` and then returns the status word (`RAN` / `SENT` / `REFUSED`). Engine console commands return the string they want shown (`FSBaseMission:consoleCommandCheatMoney` returns its messages, `FSBaseMission.lua:2504`, `:2510`), so the console likely also shows the bare status word. **Fix:** return the text instead of printing it, or return nothing; the H1a/C1a rows would need to read the status another way. What the engine does with a console command's return value is not in the references, so this is by convention only.

### Your three asks

1. **Rights check placement and the `getIsLocal()` branch: correct.** `isAdminConnection` runs first in `run`, before day validation, so a non-master with a forged count gets NOT_ADMIN, not a hint about the valid range (S6).
   - `connection:getIsServer()` refuses a connection whose far end is the server, matching `CheatMoneyEvent.lua:18` and `SaveEvent.lua:13`.
   - `getIsLocal()` is `MissionCancelEvent.lua:46`'s variant. It is safe because a listen host takes the direct `g_server` path in `requestHeatWave` and never sends the event.
   - **My mutation making every connection local: KILLED** by S2a NAMED, S2b, S2c and S6.
2. **The shared push helper: byte-equivalent.** The same bridge branch (`markFieldDirty`) and the same refresh-then-broadcast branch, now behind `g_server == nil -> return`, which the hourly site's enclosing server block already guaranteed. It is correct as moved, but see the MAJOR: the hourly call is unpinned.
3. **`onHeatResult` with the panel closed: acceptable.** `showPopup` (`CropStressSettingsPanel.lua:748-756`) only stores `popupMsg` and lines with `popupVisible = true`, so the answer shows on the next open. `self.settingsPanel` is the real manager field (`CropStressManager.lua:260`). **My mutation sending panel replies to print only: KILLED** by C3 NAMED, C4 NAMED and E1b NAMED.

### Checks

- **Events:** both readStreams end with `self:run(connection)` (the engine only calls readStream, `Server.lua:436-437`). The bench delivers every event through writeStream -> readStream, and your Q04/Q05 cover the dropped call.
- **Streams:** UInt8 days clamped 0-255 on write, re-validated on the server (`validHeatDays`, whole number 1-30). Result `Bool, String, UInt8`. Replies go only to `connection:sendEvent`, and the heat simulation's push reaches clients right away.
- **isAdmin:** `g_currentMission.isMasterUser == true`. The flag reaches a pure client through `UserDataEvent.lua:26` -> `UserManager:addMasterUser` (`:142-149`, publishes MASTERUSER_ADDED) -> `FSBaseMission:onMasterUserAdded` (`:2974-2976`). The host gets it from `addMasterUserByConnection` (`FSBaseMission.lua:463`). A second `getIsAdmin` call site does not exist.
- **Truthfulness:** a client shows only "Heat wave request sent to the server." until the reply. The refusal reads "Only a server admin can run the heat wave simulation."
- **Bench, `git archive` copy of 1c6d00f:** 2905 passed, 0 failed across 63 files (plus the `scs018_decay_catchup` copy artifact). The heat spec is 29/0, and the #191 spec 14/0.
- **TESTING row:** present.

**Not checked:** an in-game hosted or dedicated run with a real client admin and non-admin, and a re-run of your Q01-Q06.

### Commits on the branch (1)

```
1c6d00f feat(SCS #191): a client admin's heat wave runs on the host, after a server rights check
```

The title matches the commit.

Bob

PR #200, opened 2026-09-17
test(SCS #191): pin that the hourly act pushes moisture through the shared helper

TheCodingDad-TisonK:
Tests only. No production file changes.

## Why this is its own PR

Bob's review of #199 raised a MAJOR: the heat simulation and the hourly act now
share one moisture push helper, but only the heat side failed by name when that
push was removed. A mutation deleting the push from the hourly act survived.

#199 merged before this row could ride with it, so the row comes as a follow-up
on a fresh branch off development.

## What it adds

An `H hourly push` group in `SCS-191-heat_admin_request_spec_test.lua`, with a
`hourlyManager()` fixture whose integrations are all no-ops. The only thing the
rows can observe is the push the hourly act makes on the server.

| Row | Pins |
|---|---|
| H1a | reached control: the hourly body ran on the server |
| H1b NAMED | no active NetworkSync bridge: exactly one CropStressMoistureInitEvent broadcast |
| H1c | the slot refresh that must precede that snapshot ran once |
| H2a | reached control: the hourly body ran on the server |
| H2b NAMED | bridge active: markFieldDirty called exactly once |
| H2c | bridge active: nothing broadcast directly |

## Proof

Suite: 2928 assertions passed, 0 failed across 64 files.

Four mutations, each with the landed edit shown by hash and each restored with a
sha256 match afterwards:

| Mutation | Result |
|---|---|
| hourly push call removed | KILLED: H1b, H1c, H2b fail by name; H1a and H2a stay green, so the body still ran |
| bridge branch inverted | KILLED (crash kill, so it proves nothing about branch choice) |
| bridge ignored, helper always broadcasts | KILLED: H2b and H2c fail, H1 stays green, so the two cases really do separate the branches |
| refreshForPublication dropped | KILLED: H1c fails alone |

## Not in this PR

The #199 console-echo MINOR (`consoleSimulateHeat` returns a status word the
console echoes, unlike this mod's other console commands). Bob called it real by
engine convention only and cosmetic; this PR is tests only, so it stays out.

## Not checked

No in-game run. These are bench rows against the fengari mock, not a live server.


sasha-rf (approved):

Test-only, closes Bob's #199 MAJOR (the hourly-tick-calls-the-shared-push-helper gap that a mutation could remove without failing). New H group runs the real onHourlyTick end to end with every integration stubbed to a no-op, so the only observable is the push itself, and asserts it by name in both the NetworkSync-bridge-active and direct-broadcast paths (H1a-c, H2a-b). No production file changed, so the shipped zip is unaffected - matches the TESTING row's NO MANUAL TEST REQUIRED reasoning, which correctly points back to the #199 row for in-game coverage. Approved, awaiting Bob.

TheCodingDad-TisonK:
## Bob: cold diff review, PR #200 (test-only: pin the hourly moisture push)

**Head checked:** `git ls-remote` `refs/pull/200/head` = `gh` headRefOid = `662f08c`. Base `development` 2581271. One file, `tools/test/lua/SCS-191-heat_admin_request_spec_test.lua`, +58/-1. No production file is touched, so the shipped zip is development's.

**Verdict: PASS** (my #199 MAJOR is closed)

### Findings

None.

### My own mutations, on a `git archive` copy of 662f08c (each hunk confirmed, file restored and hash-compared afterwards)

- **The hourly push call removed** (`CropStressManager.lua:795` -> a comment): **KILLED**. H1b NAMED got `0:0` want `1:1`, H1c got 0 want 1, H2b NAMED got 0 want 1. H1a and H2a stayed green, so the hourly body still ran and the failure is the missing push, not a dead fixture. That is exactly the gap I raised on #199.
- **The helper's bridge branch ignored** (`if false then`, always broadcast): **KILLED** by H2b NAMED and H2c, while the H1 rows stayed green. The two cases really do separate the branches.
- **`refreshForPublication` dropped before the broadcast**: **KILLED** by H1c alone.

Bench at the head: 2924 passed, 0 failed across 64 files (plus the `scs018_decay_catchup` copy artifact), spec 35/0.

### Your two questions

1. **The skipped field-map rebuild (`lastFieldMapDay = 1`, `environment.currentDay = 1`, no `enumerateFields`): fine, and it hides nothing these rows claim.** The rebuild runs before the push and is not on the path between the hourly work and the push, so skipping it narrows the row to the one observable I asked for. What the row does not cover is the day-change rebuild itself, which was never its job; worth its own row only if someone wants that path pinned.
2. **Pinning the effect rather than the call: correct, keep it.** The rows observe a real `CropStressMoistureInitEvent` (by metatable) reaching a stubbed `g_server.broadcastEvent`, and a real `markFieldDirty`. An inlined re-implementation that still pushes once per branch is not a defect, so pinning `pushMoistureToClients` by name would test the refactor instead of the behaviour. Your judgement matches what the MAJOR asked for: something that fails by name when the client stops being told.

Your mutation 2 (bridge branch inverted) is correctly named as a crash kill that proves nothing about branch choice; mutation 3 covers that ground.

The #199 console-echo MINOR is correctly out of this PR. It stays cosmetic and unverified until someone wants it.

**Not checked:** an in-game run. These are bench rows against the fengari mock.

### Commits on the branch (1)

```
662f08c test(SCS #191): pin that the hourly act pushes moisture through the shared helper
```

The title matches the commit.

Bob

PR #202, opened 2026-09-19
fix(SCS #191): run the five dead events from readStream

TheCodingDad-TisonK:
## The defect

The engine's incoming event dispatch reads the event id, calls `readStream(streamId, connection)`, then deletes the event (`network/Server.lua:435-437`, `network/Client.lua:418-419`). **It never calls `run`.** Five of this mod's events relied on `run` being called for them, so for a remote client they arrived, were parsed, and then did nothing at all:

- `CropStressIrrigateNowEvent`
- `CropStressPivotRemoteEvent`
- `CropStressScheduleSyncEvent`
- `CropStressRainKeyCommandEvent`
- `CropStressRainKeyResultEvent`

In player terms: on a server, a joining player's Irrigate Now, pivot remote buttons, schedule Save, and rain-key commands had no effect and produced no feedback.

## The fix

`self:run(connection)` as the last line of each `readStream`. Five lines, one per event. This is the shape already used by `CropStressMoistureInitEvent:readStream` and `CropStressIrrigateNowResultEvent:readStream`. Nothing else in the five moves: no rights, ownership or reply logic changes, and no other source file is touched.

**Why this cannot double-run.** A LOCAL connection short-circuits before any serialisation and calls `event:run(self.localConnection)` directly (`network/Connection.lua:71-74`). So the local path never reaches `readStream`, and the remote path never reaches `run` on its own. That same split is why the defect survived this long: single player and a listen-server host always worked, and only a remote client's event was lost.

## Base

Cut from and rebased on nothing later than `origin/development` head **8a77fa8** (`Merge pull request #200`), re-verified unchanged immediately before the push.

## Scope limits, named deliberately

Both are pre-existing and equally true for the host before this fix, so neither is part of this defect:

1. **`CropStressPivotRemoteEvent` has no reply.** Every refusal is a server-side `print`, so a remote player whose click is refused still sees nothing. Out of scope here; worth a separate item if Desk wants the feedback.
2. **`CropStressRainKeyResultEvent` only logs.** Its `run` body writes a `devInfo` line and nothing else, so the result now arrives but still never surfaces in the UI. Out of scope here.

A sweep for any other `readStream` body in the mod without a `run` is also worth a separate item; this PR covers only the five named in the intake.

## Tests

New file `tools/test/lua/scs191_events_run_from_readstream_test.lua`, 49 assertions. Suite: **2977 passed, 0 failed across 65 files.** No existing suite or row changed.

- **One row per event** delivering through a real `writeStream` -> `readStream` round trip and asserting the run body's **effect**: the irrigate transaction is applied and a result event reaches the requester; the pivot action flips `manualMode` on the owned row; a schedule row lands on the server and the canonical row is broadcast, and separately lands on a client copy; a rain-key command mutates the system and replies with its code and new revision; a rain-key result reaches the client handler. No row calls `run` to produce the effect it measures.
- **A same-fixture CONTROL row per event** that does call `run` directly, for the single purpose of proving that fixture can reach the effect at all. So a failing row is failing on the `readStream` wiring, not on a fixture that exited early.
- **A witness delivery before each ownership refusal.** "Nothing happened" is also exactly what a dead event looks like, so each refusal row is preceded by an owned delivery on the same fixture that must visibly apply. Without this, the refusal rows would have stayed green with the fix reverted.
- **Two exactly-once rows**, one on the LOCAL path (modelling `Connection.lua:71-74`) and one on a remote delivery, so a later "helpful" `run` added to a dispatcher fails here.

### Mutation results

One mutation per event dropping its `self:run(connection)` line. Each was proved to land before its result was trusted (grep count 1 -> 0 on that file, asserted, not assumed). Each is killed by its own event's named rows, and every CONTROL row survived every mutation:

| mutation | failed rows |
|---|---|
| `CropStressIrrigateNowEvent` | `irrigate.remoteApplied`, `irrigate.remoteResultCode`, `irrigate.remoteSystemId`, `irrigate.remoteReplied`, `once.remoteResults`, `once.remoteApplied` |
| `CropStressPivotRemoteEvent` | `pivot.remoteToggled`, `pivot.refusalWitnessArrived` |
| `CropStressScheduleSyncEvent` | `schedule.serverStartHour`, `serverEndHour`, `serverManualMode`, `serverDay3`, `serverAppliedNow`, `serverBroadcast`, `serverBroadcastRow`, `clientStartHour`, `clientManualMode`, `refusalWitnessArrived`, `serverRefusedNoExtraBroadcast` |
| `CropStressRainKeyCommandEvent` | `rainKeyCmd.remoteApplied`, `remoteRevision`, `remoteReplied`, `remoteReplyCode`, `remoteReplyRevision`, `remoteRefusalReplied`, `remoteNotAuthorized` |
| `CropStressRainKeyResultEvent` | `rainKeyResult.remoteLogged`, `remoteLineNamesAction`, `remoteLineNamesRevision` |

## Reviewer notes

- **CHANGELOG entry added**, which recent fix PRs on this repo (#195, #198, #199) did not do. This one is player-visible multiplayer behaviour, so it seemed worth the line. Say the word and I will drop it.
- **Not deployed.** Five Lua lines with the syntax gate and full bench green; the dispatch said no deploy unless there was a reason and I did not find one. The zip builds cleanly. Note `build.sh` run from a worktree still writes `FS25 MODS/FS25_SeasonalCropStress.zip` rather than a worktree-scoped name (the un-merged `chore/build-sh-worktree-zip-name` gap); that zip is git-ignored and nothing was copied to the active mods folder.
- Intake and build shape: ecosystem-dev-tracking `5fd45f7` (Bob, FAST TRACK).


sasha-rf (approved):
Verified against the PR head (18b8207) and origin/development (8a77fa8): base is clean, single commit, one item.

Diff matches the description exactly: five `self:run(connection)` lines added at the end of each named readStream, nothing else touched in those files. Matches the existing pattern used by CropStressMoistureInitEvent and CropStressIrrigateNowResultEvent.

CHANGELOG `### Fixed` line: reads well, player-facing, no em dashes. Keep it in.

TESTING.md row for #202 (tracking main 6205fae): nine passes, clearly sequenced, correctly flags PASS SIX (ownership must still refuse) as the most important one. Reads clearly enough for Wizard to follow without extra context.

Approved as sasha-rf.

TheCodingDad-TisonK:
## Bob, cold diff review: PASS with MAJORs

Read cold at head `18b8207` against `origin/development` `8a77fa8`. `development` is an ancestor of the head, base is `development`, one commit on the branch.

### Engine verification, since the whole change rests on it

The premise holds and I checked it rather than took it. The receive path on both sides is emptyNew, readStream, delete, with no call to run: `network/Server.lua:435-437` and `network/Client.lua:417-419`. A local connection short-circuits before any of that, `network/Connection.lua:71-74`, `event:run(self.localConnection)`. So the five were dead only for a remote client, exactly as the PR says. Decompiled identifiers in that region are garbage, so I read the control flow and the line numbers, not the variable names.

Two further checks that were not in the PR body:

- All five call `InitEventClass` at file scope, so the ids negotiate by name and the fix is reachable.
- The sweep for other `readStream` bodies without a `run` is CLOSED for this repo. All 14 event files in `src/events/` that define `readStream` now call `self:run` exactly once, and there is no `readStream` outside `src/`. The sweep stays open across the other mods.

Replying from inside `readStream` is safe here: `Connection:sendEvent` writes through the write offset while the engine's post-readStream validation at `network/Server.lua:443-444` reads only the read offset, and nine sibling events in this repo already reply this way.

### Findings

`tools/test/lua/scs191_events_run_from_readstream_test.lua:315`: MAJOR: the only conditional read in all five readStream bodies is RainKeyCommand's `hasValue` branch at `src/events/CropStressRainKeyCommandEvent.lua:43-44`, and every row delivers `FIT` with `value=nil`, so the Float32 leg is never taken. `wireClean` cannot see a branch it never enters, and a branching read is exactly where a write/read asymmetry hides. fix: add one `SET_TRIP_MM` delivery carrying a numeric value.

`tools/test/lua/scs191_events_run_from_readstream_test.lua:224`: MAJOR: the `broadcastEvent` stub drops its second argument, so nothing pins the `false` at `src/events/CropStressScheduleSyncEvent.lua:179` and `src/events/CropStressPivotRemoteEvent.lua:146`. Per `network/Server.lua:542-544` that argument is the listen-host gate: flipped to true, the host's own client would receive its own broadcast and apply twice, and every row in this file would stay green. This is in the blast radius because the change is what makes these events live. fix: capture the second argument and assert it is false.

`ecosystem-dev-tracking TESTING.md:66`: MAJOR: the watch list names "Invalid event id" and "Not all bits read in event" but not the one error this shape produces. `network/Server.lua:440` logs "Could not validate read bits because stream was already closed by event", which is the engine's own guard against an event that replies while its read stream is open. That is the highest-consequence unknown in this change and the row should name the string a tester greps for. fix: add it to the watch list.

`tools/test/lua/prelude.lua:88-89`: MINOR: `streamWriteUIntN` and `streamReadUIntN` discard the width argument, so `wireClean` is blind to UIntN width drift. None of the five uses UIntN, so the claim is honest here, but the rows read as a general wire guarantee and are not one. fix: one comment scoping `wireClean` to the typed lanes.

`tools/test/lua/scs191_events_run_from_readstream_test.lua:385`: MINOR: `once.localRequests` cannot fail under any mutation of this diff, since no run body re-sends the request class. It is a future-regression pin, not a detector. fix: none needed, just do not count it toward coverage.

`tools/test/lua/scs191_events_run_from_readstream_test.lua:386`: MINOR: the comment says one result "goes back to the requester", but on the local path the result is sent and then dropped by its own guard at `src/events/CropStressIrrigateNowResultEvent.lua:66`. Accurate to the engine, over-claimed in the label. fix: reword to "is sent exactly once".

### The four questions asked in the handoff

1. **Is `deliverRemote` honest?** Yes. It is `Server.lua:435-437` and `Client.lua:417-419` with the run deliberately absent. It omits two things, neither of which weakens the claim: the post-readStream bits validation, which touches only the read offset, and the fact that the engine passes the receiving side's connection object rather than the sender's. None of the five `writeStream` bodies reads `connection` at all, it appears only in their signatures, so reusing one object is sound.

2. **The exactly-once pair.** The model is right and matches `Connection.lua:71-74` verbatim. The half it does not cover is the other direction of the same guarantee, `Server:broadcastEvent`'s `sendLocal` gate. That is MAJOR 2 above.

3. **Is the RainKeyResult log line acceptable as an effect?** Yes, with the claim scoped. It is the only observable that run body has, so asserting it is the honest maximum, and with the fix reverted the row goes red, which is what a detector has to do. Say plainly that it witnesses dispatch and not behaviour: it proves the event now reaches run, and proves nothing about a player seeing a result. The TESTING row already gets this right.

4. **The two judgment calls.** Both right. The CHANGELOG `### Fixed` belongs there: the section did not exist under `[Unreleased]`, and I swept the whole block rather than the one line, since a later PR can falsify an existing entry. Nothing in it contradicts or duplicates the new bullet. `#195`, `#198` and `#199` omitting an entry is a gap in those PRs, not a precedent. Not deploying is also right: every pass on the TESTING row needs a joined client, and a solo run passes whether or not the fix is present, so a local deploy would produce a green that means nothing.

### Scope limits, agreed as stated

PivotRemote has no reply, so a refused remote click still shows a joined player nothing, and RainKeyResult still only logs. Both were true before this change. One to add to the in-game pass rather than the code: an ACCEPTED pivot action applies on the server, and whether the joined player then sees the arm move is the Reinke placeable's own replication, not ours. If the pivot visibly moves for the host but not the client, that is a separate defect and not a regression from this PR.

### Verdict

**PASS with MAJORs.** The source change is five lines and each one is correct. Every run body is guarded for the side it belongs to: IrrigateNow, PivotRemote and RainKeyCommand are server-only, ScheduleSync branches on `g_server` and applies read-only on a client, RainKeyResult only logs. Streams are symmetric in all five. The three MAJORs are all bench and test-instruction gaps, not defects in the shipped change, and none of them needs to block the merge if they land in a follow-up.

Commits on the branch, for the title check:

- `18b8207` fix(SCS #191): run the five dead events from readStream

One commit, and the PR title matches it.

-- Bob

TheCodingDad-TisonK:
## Bob, re-review at 4320c53: PASS, no new findings

Test-only, one file, source untouched. All three MAJORs closed correctly and the three MINORs taken as scope corrections to how the rows read.

- MAJOR 2 is closed the right way round. Both stubs now capture arg 2 and assert `false`, and the flip test showing only `pivot.broadcastNotSentLocal` and `schedule.broadcastNotSentLocal` fail is the evidence that matters: it proves the new rows are detectors rather than decoration, and it proves the old file could not see that argument at all.
- MAJOR 1 is closed with a positive assertion on a real value, `7.5` surviving the wire and landing on the dial, so the row cannot pass on an early exit the way a refusal row can.
- MAJOR 3 names the exact string and tells the tester to report it even when the passes above look green, which is the part that makes it useful.

Checked the surrounding code rather than only the delta, since a fix round only proves its changed lines. `pivotFixture` gained a second return value and both older call sites take only the first, which is harmless. `scheduleFixture`'s client branch leaves `seen.sendLocal` empty and no row asserts it there. The new `SET_TRIP_MM` block builds its own fixture and connection, so it does not disturb the unauthorized row it now sits above. The four line citations added in comments all land: `Server.lua:440`, `prelude.lua:88-89`, `CropStressRainKeyCommandEvent.lua:43-44` and `CropStressIrrigateNowResultEvent.lua:66`.

Verdict stands at **PASS**. This head carries no BLOCKERs and no MAJORs.

Note for the merge: the approval on `18b8207` does not cover this head. An old approval never covers code it did not see, so this needs a fresh one before merge.

-- Bob

sasha-rf (approved):
Re-reviewed at new head 4320c53 (my prior approval was against 18b8207, now stale).

Verified independently:
- The five source `self:run(connection)` lines are byte-identical to what I reviewed before (same blob hashes in the diff against origin/development).
- The second commit (4320c53) touches only tools/test/lua/scs191_events_run_from_readstream_test.lua. CHANGELOG.md is untouched.
- Bob's three MAJORs (unentered RainKeyCommand branch, missing broadcastEvent second-arg assertion for the listen-host double-apply gate, and the over-claiming row labels) are all closed in that commit; no BLOCKERs, nothing touching the five source lines.

TESTING.md row (tracking main 7372290): re-read. The added watch-for-this-string line ("Could not validate read bits because stream was already closed by event", Server.lua:440) is a good catch and correctly tells the tester to report it even on an apparent PASS. Bob added to the reviewer column. Still clear for Wizard to follow.

Approved as sasha-rf at 4320c53.

